PR #2, opened 2024-08-22
feat: Configure EventKit binary to support multiple service with graceful shutdown

alexluong:
- [x] Configure EventKit binary to support multiple service
- [x] Graceful shutdown
- [x] Production Dockerfile to build release image
- [x] Example with custom YAML config

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/redis/go-redis/v9 v9.6.1
 	github.com/spf13/viper v1.19.0
-	golang.org/x/sync v0.8.0
 )
 
 require (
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -2,32 +2,47 @@ package config
...
+	Port int
 
-	RedisHost     = ""
-	RedisPort     = ""
-	RedisPassword = ""
-	RedisDatabase = 0
+	RedisHost     string
+	RedisPort     string
+	RedisPassword string
+	RedisDatabase int
```
alexbouchardd: We should still support default values

internal/services/data/data.go
```diff
@@ -0,0 +1,43 @@
...
+		<-ctx.Done()
+		log.Println("shutting down data service")
+	}()
+	return &DataService{}
+}
+
+func (s *DataService) Run(ctx context.Context) error {
+	log.Println("running data service")
+
+	if os.Getenv("DISABLED") == "true" {
```
alexbouchardd, line 28: Unclear what DISABLED here refers to?

alexluong: Ah this is mainly just a temporary setup. I have this data service logging the destination count every minute and it gets noisy so I added this mechanism to suppress it. Will remove it once we implement it.

I added the explanation in `docker-compose.yml` but forgot to also add it here, my bad 😅

PR #4, opened 2024-08-23
feat: Set up basic OTEL with development configuration

alexluong:


cmd/app/main.go
```diff
@@ -29,25 +30,46 @@ func main() {
...
+		}
+		// Handle shutdown properly so nothing leaks.
+		defer func() {
+			err = errors.Join(err, otelShutdown(context.Background()))
+		}()
+
+		// QUESTION: what if a service doesn't need Redis? Is it unnecessary to initalize the client here?
+		if err := redis.InstrumentOpenTelemetry(); err != nil {
+			return err
+		}
```
alexbouchardd: Yes, the log service wouldn't use it. I'm not sure why we need Redis telemetry in the first place?

In any case we could skip if the service is `log`

cmd/app/main.go
```diff
@@ -29,25 +30,46 @@ func main() {
...
+	// Construct services based on config
 	services := []Service{}
-
-	switch flags.Service {
-	case "api":
+	switch config.Service {
+	case config.ServiceTypeAPI:
 		services = append(services, api.NewService(ctx, wg))
-	case "data":
+	case config.ServiceTypeData:
```
alexbouchardd: Service renamed to `Log`

cmd/app/main.go
alexluong: This is the Redis's client telemetry, not the telemetry of the Redis cluster itself. This adds data like how long a command took. For example:

CleanShot 2024-08-24 at 21 06 39

> In any case we could skip if the service is log

Got it, let's add this as a TODO since I'm not 100% sure about the pattern around Redis & Config yet. I'm gonna get around to this when we set up testing. The concern is around dependency chain vs accepting values to the init function for inversion of control.

cmd/app/main.go
```diff
@@ -23,58 +26,87 @@ type Service interface {
...
+		// Handle shutdown properly so nothing leaks.
+		defer func() {
+			err = errors.Join(err, otelShutdown(context.Background()))
+		}()
+
+		// QUESTION: what if a service doesn't need Redis? Is it unnecessary to initalize the client here?
+		if err := redis.InstrumentOpenTelemetry(); err != nil {
+			cancel()
+			return err
+		}
```
alexbouchardd, line 69: I wouldn't have Redis OTEL from the application side, and I think it's beyond the scope of responsibility. cc @mkherlakian Do you agree?

configs/dev/uptrace/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+      POSTGRES_DB: uptrace
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U uptrace -d uptrace"]
+      interval: 1s
+      timeout: 1s
+      retries: 30
+    volumes:
+      - "pg_data3:/var/lib/postgresql/data/pgdata"
+    ports:
+      - "5432:5432"
```
alexbouchardd, line 36: Suprised you need both PG and CLickhouse

internal/otel/otel.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"go.opentelemetry.io/contrib/instrumentation/host"
+	"go.opentelemetry.io/contrib/instrumentation/runtime"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/log/global"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+// TODO: Consider supporting the official OTEL configuration format.
+// https://opentelemetry.io/docs/collector/configuration/
```
alexbouchardd, line 16: Why not?

internal/redis/redis.go
```diff
@@ -1,29 +1,53 @@
...
+	}
+	if config.OpenTelemetry.Metrics != nil {
+		if err := redisotel.InstrumentMetrics(client); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
```
alexbouchardd, line 41: See redis otel comment

cmd/app/main.go
alexbouchardd: That seems like too much granularity for a user that's going to deploy the service

cmd/app/main.go
alexluong: Maybe I'm missing the point of OTEL in the design spec.

This OTEL setup is to set up observability for EventKit services. How long each request may take, broken down by REDIS call, enqueuing messages, etc.

Maybe when you say OTEL you're referring to specific EventKit business logic that you want to expose to users, like how long an event moves through the system.

Personally, I think this provides helpful observability and if people don't care they can use the config to get the data they want?

configs/dev/uptrace/docker-compose.yml
alexluong: Yes, Uptrace uses PG for metadata and CH for the telemetry data itself.

internal/otel/otel.go
alexluong: It actually does already support natively because of the Go SDK, but there's a slight difference in behavior around the exporter endpoints because I wanted to let people configure what they care about. We can write the configuration docs in more details in the deployment documentation.

cmd/app/main.go
alexbouchardd: The design intention is that EvenKit will produce its own OTEL metrics. If we export traces for the underlying implementation, that is a good question. Redis metrics would be much more useful to a maintainer than an end user. Let me get back to you on this. We can merge in the meantime

alexluong: Got it, and yes this is purely for the maintainer.

As for its own OTEL metrics, it sounds like EventKit will produce the telemetry itself and then the log service (or another service) will be the BE to store & serve the data to the end user. That sounds like a totally different scope tho.

PR #6, opened 2024-08-28
feat: API key authentication middleware

alexluong:


internal/services/api/auth_middleware.go
```diff
@@ -0,0 +1,41 @@
...
+			// Currently we don't have clear specs on how to send back error message.
+			c.AbortWithStatus(http.StatusUnauthorized)
+			return
+		}
+		if authorizationToken != apiKey {
+			// TODO: Consider sending a more detailed error message.
+			// Currently we don't have clear specs on how to send back error message.
+			c.AbortWithStatus(http.StatusUnauthorized)
+			return
+		}
```
alexluong, line 31: I want to point this out as the spec hasn't mentioned errors yet, afaik. Do you want to circle back on this later?

Happy to sort this out now if you can share what you have in mind or provide the OpenAPI schema spec or something like that.

internal/services/api/auth_middleware_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+)
+
+func setupRouter(apiKey string) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	r := gin.Default()
+	r.Use(apiKeyAuthMiddleware(apiKey))
```
alexluong, line 15: Sooo in the spec I saw you mentioned JWT token but I'm not super super clear on that just yet.

My understanding:
- optional API key auth mechanism for internal purposes
- JWT auth mechanism for portal / per-tenant usage

Please correct me if there's any misunderstanding here. Will tackle the Tenant API now and implement the JWT mechanism in another PR.

alexbouchardd: That's correct, we'll need to think through how that works since the tenant is implied with the JWT really the endpoint wouldn't be have `/:tenant_id`

alexluong: It's more of a API design concern, because if you'd like to remove the `/:tenant_id` path param we can get the tenant ID from the JWT tokentoo.

PR #7, opened 2024-08-28
feat: Tenant API & JWT authentication

alexluong:
This is a fairly loaded PR. Here are everything it does:

- [x] Tenant's CRUD based on spec (with just ID & created at timestamp for now)
- [x] JWT generation & verification
- [x] JWT middleware mechanism
- [x] Refactor router to have 2 different authentication scopes: admin-only and tenant-scope

Added tests for a lot of the new code here. Just a sidenote, I really do enjoy writing tests in Go and the workflow so far. Caught a bug in the auth middleware from writing a more semi-comprehensive test suite.

### DIFF
--- a/internal/destination/handlers_test.go
+++ b/internal/destination/handlers_test.go
@@ -1,0 +1,214 @@
+package destination_test
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/hookdeck/EventKit/internal/destination"
+	"github.com/hookdeck/EventKit/internal/util/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func setupRouter(destinationHandlers *destination.DestinationHandlers) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	r := gin.Default()
+	r.GET("/destinations", destinationHandlers.List)
+	r.POST("/destinations", destinationHandlers.Create)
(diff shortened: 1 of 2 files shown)


internal/services/api/router_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"github.com/google/uuid"
+	"github.com/hookdeck/EventKit/internal/destination"
+	"github.com/hookdeck/EventKit/internal/services/api"
+	"github.com/hookdeck/EventKit/internal/tenant"
+	"github.com/hookdeck/EventKit/internal/util/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRouterWithAPIKey(t *testing.T) {
+	t.Parallel()
```
alexluong, line 17: Can you go over this test suite (this file)? I'm trying to test the router with a strong emphasis on the auth mechanism. Some questions:

1. Can you follow the thought process / auth mechanism logic here? Is this the right approach? Any case I'm missing here?
2. Do you feel this is too "auth-focused" and should be moved to `auth_middleware_test.go` instead? In that case, we can set up a smaller router vs this router with tenant & destination routes. Personally I'm fine with this and continue moving forward and we can consider that refactor after but curious if you share the sentiment.

internal/config/config.go
```diff
@@ -16,10 +16,11 @@ const (
...
 type Config struct {
-	Service  ServiceType
-	Port     int
-	Hostname string
-	APIKey   string
+	Service   ServiceType
+	Port      int
+	Hostname  string
+	APIKey    string
+	JWTSecret string
```
alexluong, line 23: I assume we would need to add a `JWT_SECRET` env here, correct?

Issue #39, opened 2024-10-03
Refactor persistence layer


alexluong:
When first implementing the tenant & destination logic with Redis, I introduced 2 model structs `TenantModel` and `DestinationModel`. They were originally stateful with the Redis client, but as things progress, they have become stateless because I wanted the option of passing a Redis transaction to each individual operations.

Then, with ClickHouse, we also introduce `EventModel` and `DeliveryModel` where they're also stateless and accept a ClickHouse connection.

Even though they're stateless, they're still being passed around in the code as if they're stateful. While passing them around is a common thing in Golang, I do think having multiple models like this is a bit clunky.

I think we should consider refactoring to either of this option:

1: a singular "Model" interface that can work with every resource types.
2: two interfaces for MetadataModel and EventModel (or MetadataRepo and EventRepo) -- splitting via domain but also different storage solution

We should also consider making them stateful again and designing the operations so that the transaction can be moved into this layer instead of having the transaction managed from the outer layer.

One interesting read regarding this topic: https://threedots.tech/post/database-transactions-in-go/

Some other perks of these models being interfaces are
- mockable during tests
- limit the impacted surface if we ever want to support different storage type

---

Here are all the operations we have (or will have):

Redis:
- retrieve tenant
- set tenant
- delete tenant
- list destination by tenant
- retrieve destination
- set destination
- delete destination

ClickHouse
- list event
- retrieve event
- insert many event
- list delivery
- retrieve delivery
- insert many delivery

PR #41, opened 2024-10-04
refactor: Use metadata repo interface over destination & tenant models

alexluong:
implements #39


internal/deliverymq/messagehandler.go
```diff
@@ -18,23 +18,26 @@ import (
...
 }
 
 var _ consumer.MessageHandler = (*messageHandler)(nil)
 
-func NewMessageHandler(logger *otelzap.Logger, redisClient *redis.Client, destinationModel *models.DestinationModel, logMQ *logmq.LogMQ) consumer.MessageHandler {
+func NewMessageHandler(
+	logger *otelzap.Logger,
+	redisClient *redis.Client,
+	logMQ *logmq.LogMQ,
+	metadataRepo models.MetadataRepo,
```
alexbouchardd, line 34: I realize that I'm not sure I understood you previously. Unclear to me what "MetadataRepo" is supposed to be

alexluong: So we have 2 types of data persistence:

1: events-related data in ClickHouse -- LogRepo
2: tenant and destination data in Redis -- MetadataRepo

Hope that clarifies what they mean. Let me know if you prefer to rename to something that may make more sense to you.

alexbouchardd: But tenant and destination are still distinct models no?

alexluong: That's what I had at the beginning but I think the DX is a bit clunky. The idea is not to group things by the model (or table) but rather a domain (related models).

For example, with the idea that when we add a new destination, we will also need to update a tenant-related value. That's why I think it makes more sense to group them together (metadata repo, or whatever you'd like to call it) instead of having a tenant model + destination model.

alexbouchardd: I see, call it something like `Entities` "metadata" is confusing since we already use that term in the event context and to me a log is metadata of the delivery

alexluong: So the terminology shall be `EntityRepo` & `LogRepo`? Or `Entities` and something else?

As you can see in the code, but just to share some snippets as examples of how it's used

```
metadataRepo.RetrieveDestination()
metadataRepo.UpsertTenant()
logRepo.InsertManyEvent()
// etc
```

alexbouchardd: `entityStore`
`logStore`

Perharps?

alexluong: @alexbouchardd, sounds good to me, will update them soon!

I have another PR #44 that touches on this model layer quite a bit. To avoid potential git conflict issues, what do you think if we merge this and that PR (still WIP) and then I'll open a separate PR to rename as suggested?

alexbouchardd: Yeah all good

PR #61, opened 2024-10-23
feat: Use retry scheduler for deliverymq

alexluong:
This is the first implementation of a Redis-based retry scheduler for DeliveryMQ.

### Scheduler

I implemented a simple Scheduler interface:

```go
type Scheduler interface {
	Init(context.Context) error
	Schedule(context.Context, string, time.Duration) error
	Monitor(context.Context) error
	Shutdown() error
}

func New(..., exec func(context.Context, string) error) {}
```

`Init()` and `Shutdown()` are operational functions. The main interface is the `Schedule()` and `Monitor()` functions as well as the `New()` constructor that accepts an `exec` function.

Here's how to use the scheduler:

1: create a new scheduler with an exec function to run when something is due
2: schedule things to run
3: monitor the scheduler in another process

Under the hood, it uses [`semihbkgr/go-rsmq`](https://github.com/semihbkgr/go-rsmq) which is a port of [`smrchy/rsmq`](https://github.com/smrchy/rsmq).

### DeliveryMQ retry scheduler

We use the scheduler in the deliverymq handler instead of nacking messages when 1: there's an error when deliverying (publishing) event to the destination && 2: the event is eligible for retry.

This is the data we store in Redis for each retry message:

```go
type RetryMessage struct {
	DeliveryEventID string
	EventID         string
	TenantID        string
	DestinationID   string
	Telemetry       *models.DeliveryEventTelemetry
}
```

Once it's time, we turn this `RetryMessage` back into delivery event (without event data) and move it back to the deliverymq.

We can optimize this in the future once we improve the log store to support our specific query pattern. For now, the event data is stored in CH and we'll query it via `tenant_id` & `event_id`. As mentioned, this may change in the future as we continue working on the log service and CH schema.

Note: currently, the retry delay is 1s after every attempt. This issue #62 is to implement exponential retry logic instead of the linear 1s delay here.

### DIFF
--- a/internal/deliverymq/messagehandler.go
+++ b/internal/deliverymq/messagehandler.go
@@ -56,13 +56,10 @@ func NewMessageHandler(
 func (h *messageHandler) Handle(ctx context.Context, msg *mqs.Message) error {
 	deliveryEvent := models.DeliveryEvent{}
 	if err := deliveryEvent.FromMessage(msg); err != nil {
-		// TODO: question - should we ack this instead?
-		// Since we can't parse the message, we won't be able to handle it when retrying later either.
 		msg.Nack()
 		return err
 	}
 	if err := h.ensureDeliveryEvent(ctx, &deliveryEvent); err != nil {
-		// Question: nack or retryScheduler.Schedule?
 		msg.Nack()
 		return err
 	}


internal/deliverymq/messagehandler.go
```diff
@@ -48,25 +55,34 @@ func NewMessageHandler(
...
-	err := deliveryEvent.FromMessage(msg)
-	if err != nil {
+	if err := deliveryEvent.FromMessage(msg); err != nil {
 		// TODO: question - should we ack this instead?
 		// Since we can't parse the message, we won't be able to handle it when retrying later either.
 		msg.Nack()
 		return err
 	}
+	if err := h.ensureDeliveryEvent(ctx, &deliveryEvent); err != nil {
+		// Question: nack or retryScheduler.Schedule?
```
alexluong: This occurs when the race condition happens when the retry happens before event data is persisted.

It could also happen because of systemic error (unable to connect to ClickHouse, etc).

Should we nack or attempt another retryScheduler.Schedule?

alexbouchardd: We should nack and retry later

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (s *schedulerImpl) Init(ctx context.Context) error {
+	if err := s.rsmqClient.CreateQueue(s.name, s.config.visibilityTimeout, rsmq.UnsetDelay, rsmq.UnsetMaxsize); err != nil && err != rsmq.ErrQueueExists {
+		return err
+	}
+	return nil
+}
+
+func (s *schedulerImpl) Schedule(ctx context.Context, message string, delay time.Duration) error {
+	if _, err := s.rsmqClient.SendMessage(s.name, message, uint(delay.Seconds())); err != nil {
```
alexbouchardd, line 66: Edge case here, but in the case of redelivery, you can end up with multiple messages scheduled; we can handle this by setting the message ID to something deterministic like the event_delivery_id. Within Hookdeck we forked RMSQ to add the ability to pass a message ID I believe

alexluong: Can you clarify this edge case a bit?

1: how can multiple messages be scheduled? When the server crash between a retry scheduled & message acking?
2: what do we do with the event_delivery_id to deduplicate? Is that on the retry monitor logic or the delivery message handler?

alexbouchardd: > How can multiple messages be scheduled? When the server crash between a retry scheduled & message backing?

For example yes

> what do we do with the event_delivery_id to deduplicate?

Not sure to be following the question but if the message that you push into Redis has a deterministic ID then you'd be overriding the existing sorted set member instead of adding a new one making it idempotent

alexluong: Ah I see. Is there still a case where the 2nd message is sent AFTER the retry is already handled, so essentially the item would be added to sorted set, pop, then added again?

alexbouchardd: Yeah it's possible although depending on the retry delay might not be very likely

internal/deliverymq/messagehandler.go
alexluong: Alright, confirming we'll nack and let the mq retry whenever. Because of that, it will NOT use the Redis-based exponential retry schedule. Just wanna be clear here.

internal/scheduler/scheduler.go
alexluong: Created #64 to track this issue.

PR #69, opened 2024-10-28
feat: Max retry count

alexluong:
implements #66

### DIFF
--- a/internal/deliverymq/messagehandler.go
+++ b/internal/deliverymq/messagehandler.go
@@ -140,6 +140,11 @@ func (h *messageHandler) doHandle(ctx context.Context, deliveryEvent models.Deli
 // - OR if the error is a publish error + event is eligible for retried + the attempt is less than max retry --> should be retried
 // Clarification between attemp & max retry: Event.Attempt is zero-based numbering. If Event.Attempt is 4, that means the event has failed 5 times.
 // That means 1 original attempt + 4 retries. If max retry is 5, then it should be retried 1 more time. Total attempt will be 6.
+//
+// Question: what to do if there's an "internal error"? How should that count against the attempt count?
+// For example, what if the error happens AFTER deliverying the message (doesn't matter whether it's successful or not),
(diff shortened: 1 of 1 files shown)


internal/deliverymq/messagehandler.go
```diff
@@ -132,6 +134,25 @@ func (h *messageHandler) doHandle(ctx context.Context, deliveryEvent models.Deli
...
+// Qualifications:
+// - if the error is an internal error --> should be retried
+// - OR if the error is a publish error + event is eligible for retried + the attempt is less than max retry --> should be retried
+// Clarification between attemp & max retry: Event.Attempt is zero-based numbering. If Event.Attempt is 4, that means the event has failed 5 times.
+// That means 1 original attempt + 4 retries. If max retry is 5, then it should be retried 1 more time. Total attempt will be 6.
+//
+// Question: what to do if there's an "internal error"? How should that count against the attempt count?
+// For example, what if the error happens AFTER deliverying the message (doesn't matter whether it's successful or not),
+// say logmq.Publish fails. Should that count as an attempt? What about an error BEFORE deliverying the message?
+// Should we write code to differentiate between these two types of errors (predeliveryErr and postdeliveryErr, for example)?
```
alexluong, line 147: Wanna highlight this question regarding error handling

alexbouchardd: In neither case it should count as an attempt, for an attempt to be valid, it needs to have succeeded through out the full flow

alexbouchardd: So really in case of any of those issues the message is nacked and it's the same attempt that's redelivered

alexluong: Okay, just to summarize to make sure I understand how to handle this correctly then. We have 2 mechanism for retry:

1: nack and the mq will retry whenever
2: schedule a retry via RSMQ

We only use 2 when the retry is for a failed delivery attempt. All other cases, we nack.

Does that sound right?

PR #86, opened 2024-11-21
example: Quick start Docker Compose setup

alexluong:
Steps:

```sh
$ cd examples/docker-compose
$ docker-compose -f compose.yml -f compose-aws.yml up -d
```

This should give you a full Outpost environment. You can then follow the [phase 2 of the guide](https://github.com/hookdeck/outpost/blob/main/docs/contributing/step-by-step.md#phase-2-create-tenant--destination) to see Outpost in action.

We can consider Dockerize the mock servers and make the DX and the quickstart experience better so that users don't have to start multiple running services to test out Outpost. I consider that a separate scope that we can consider improving later.

---

In the future, we can have other options besides `compose-aws.yml`:

- `compose-rabbitmq` or `compose-gcppubsub`, etc. for other mqs
- `compose-otel` for OpenTelemetry
- `compose-destinations.....` for mock destination server as mentioned above

The goal is for users who want to give Outpost a try can simply run a `docker-compose ... up` and can start playing aroud with the service right away.

### DIFF
--- a/examples/docker-compose/compose.yml
+++ b/examples/docker-compose/compose.yml
@@ -1,7 +1,7 @@
 name: outpost-example
 services:
   api:
-    image: hookdeck/outpost:latest-arm64
+    image: hookdeck/outpost:latest
     env_file: .env
     depends_on:
       redis:
@@ -14,7 +14,7 @@ services:
       SERVICE: api
 
   delivery:
-    image: hookdeck/outpost:latest-arm64
+    image: hookdeck/outpost:latest
     env_file: .env
     depends_on:
       redis:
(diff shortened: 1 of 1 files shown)


examples/docker-compose/compose.yml
```diff
@@ -0,0 +1,93 @@
+name: outpost-example
+services:
+  api:
+    image: hookdeck/outpost:latest-arm64
```
alexluong: This image assumes either Outpost image is already published to Docker or you have a local image running. Let me know if you'd like to give this a try locally.

The short version is:
- install `goreleaser` cli
- at root, run `$ goreleaser release -f ./build/.goreleaser.yaml --snapshot --clean` to build the local Docker image.

Issue #15, opened 2024-09-15
Implement destination types


alexluong:
Planned

- [x] Webhooks
- [ ] Hookdeck Event Gateway
- [x] AWS SQS
- [ ] AWS SNS
- [ ] AWS EventBridge
- [ ] GCP PubSub
- [x] RabbitMQ
- [ ] Kafka

Others

- [ ] Azure Service Bus
- [ ] Azure EventGrid
- [ ] MQTT
- [ ] Segment & other CDPs
- [ ] Zapier & other no code tools
- [ ] S3 / GCS / Azure Blob Storage
- [ ] NATS

PR #117, opened 2024-11-28
docs: improve README / add quickstart

leggetter:


README.md
```diff
@@ -1,5 +1,123 @@
...
+[![License](https://img.shields.io/badge/License-Apache--2.0-blue)](#license)
+[![Go Report Card](https://goreportcard.com/badge/github.com/hookdeck/outpost)](https://goreportcard.com/report/github.com/hookdeck/outpost)
+[![Issues - Outpost](https://img.shields.io/github/issues/hookdeck/outposta)](https://github.com/hookdeck/outpost/issues)
+![GitHub Release](https://img.shields.io/github/v/release/hookdeck/outpost)
+
+</div>
+
+# Open Source Event Destinations Infrastructure
+
+Outpost is a self-hosted and open-source product that enables event producers to add Event Destinations to their platform with support for destination types such as Webhooks, Hookdeck Event Gateway, AWS SQS, Amazon EventBridge, GCP Pub/Sub, MQTT, and Azure EventBus.
```
alexluong: Not sure MQTT is the right thing here in our case. If you mean RabbitMQ, we support AMQP 0.9.1 instead of MQTT. Do keep in mind AMQP is a bit awkward, there are other standards like AMQP 1.0 which is a different protocol, so the versioning is important here.

README.md
```diff
@@ -1,5 +1,123 @@
...
+```
+
+Wait until you get a `OK%` response.
+
+Create a tenant:
+
+```sh
+curl --location --request PUT 'localhost:3333/api/v1/<TENANT_ID>' \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Bearer <API_KEY>'
```
alexluong, line 87: Should we specify the API_KEY here for folks to copy as the `.env.example` has a default api key value?

README.md
```diff
@@ -1,5 +1,123 @@
...
+curl --location 'localhost:3333/api/v1/<TENANT_ID>/destinations' \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Bearer <API_KEY>' \
+--data '{
+    "type": "webhook",
+    "topics": ["*"],
+    "config": {
+        "url": "<URL>"
+    },
+    "credentials": {}
```
alexluong: Just a note, I recalled you mentioned the need for `credentials` here. It's no longer required, so there's that. No problem keeping it there tho.

README.md
```diff
@@ -1,5 +1,123 @@
...
+```
+
+Publish an event:
+
+```sh
+curl --location 'localhost:3333/api/v1/publish' \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Bearer <API_KEY>' \
+--data '{
+    "tenant_id": "<TENANT_ID>",
```
alexluong, line 114: Shall we specify that they should replace the tenant ID value here? Just wondering since if they don't pay attention they may miss it. All good regardless tho

README.md
leggetter: What destination types do we plan to support for EA?

What destinations types are on the roadmap beyond that?

alexluong: This [issue](https://github.com/hookdeck/outpost/issues/15) tracks the answer to your questions here. You can also find them in the spec.

Currently, we only have implementation for RabbitMQ (AMQP 0.9.1) & AWS SQS. We probably want GCP PubSub & maybe some others (Azure, Kafka, etc.) for EA, if there's a need and it works with our timeline.

Long term, I believe the goal is to support all destination types as long as it makes sense to be integrated with Outpost.

README.md
```diff
@@ -1,5 +1,124 @@
...
+[![License](https://img.shields.io/badge/License-Apache--2.0-blue)](#license)
+[![Go Report Card](https://goreportcard.com/badge/github.com/hookdeck/outpost)](https://goreportcard.com/report/github.com/hookdeck/outpost)
+[![Issues - Outpost](https://img.shields.io/github/issues/hookdeck/outposta)](https://github.com/hookdeck/outpost/issues)
+![GitHub Release](https://img.shields.io/github/v/release/hookdeck/outpost)
+
+</div>
+
+# Open Source Event Destinations Infrastructure
+
+Outpost is a self-hosted and open-source product that enables event producers to add Event Destinations to their platform with support for destination types such as Webhooks, Hookdeck Event Gateway, AWS SQS, Amazon EventBridge, GCP Pub/Sub, MQTT, and Azure EventBus.
```
leggetter:
```suggestion
Outpost is a self-hosted and open-source product that enables event producers to add Event Destinations to their platform with support for destination types such as Webhooks, Hookdeck Event Gateway, Amazon EventBridge, AWS SQS, AWS SNS, GCP Pub/Sub, RabbitMQ, and Kafka.
```

PR #158, opened 2024-12-19
feat: Add working destination settings

alexbouchardd:
> [!IMPORTANT]
> Enhances destination settings with new configuration fields, enable/disable and delete actions, and updates UI components and styles for improved functionality.
> 
>   - **Behavior**:
>     - Adds `DestinationConfigFields` component in `DestinationSettings.tsx` for rendering configuration and credential fields.
>     - Implements enable/disable and delete functionality for destinations in `DestinationSettings.tsx`.
>     - Adds loading states for buttons in `Button.tsx` and `DestinationSettings.tsx`.
>     - Displays loading spinner in `Destination.tsx` when destination data is not yet available.
>   - **Models**:
>     - Introduces `Destination`, `ConfigField`, `CredentialField`, and `DestinationTypeReference` types in `Destination.ts`.
>   - **UI Components**:
>     - Updates `Button` component to support `loading` and `danger` states in `Button.tsx` and `Button.scss`.
>     - Adds `CopyButton` component for copying text to clipboard in `CopyButton.tsx`.
>     - Adds new icons `DisableIcon` and `DeleteIcon` in `Icons.tsx`.
>   - **Styles**:
>     - Adds styles for `DestinationConfigFields` in `DestinationConfigFields.scss`.
>     - Updates styles for `DestinationSettings` in `DestinationSettings.scss`.
>     - Adjusts button styles in `Button.scss` to accommodate new states.
>   - **Misc**:
>     - Changes `type` from `string` to `text` in `core.json` files for AWS SQS, RabbitMQ, and Webhook.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=hookdeck%2Foutpost&utm_source=github&utm_medium=referral)<sup> for 1cfc9570ba9b10a5dbea08fff0e93fa715bebed6. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/internal/portal/src/typings/Destination.ts
+++ b/internal/portal/src/typings/Destination.ts
@@ -1,5 +1,3 @@
-type DestinationType = "rabbitmq" | "aws_sqs" | "webhook";
-
 interface ConfigField {
   type: "text";
   label: string;
@@ -24,13 +22,11 @@ interface DestinationTypeReference {
 
 interface Destination {
   id: string;
-  type: DestinationType;
+  type: string;
   config: Record<string, any>;
   credentials: Record<string, any>;
   label: string;
   description: string;
-  icon: string;
-  instructions: string;
   target: string;
   disabled_at: string;
 }


internal/portal/src/typings/Destination.ts
```diff
@@ -0,0 +1,43 @@
+type DestinationType = "rabbitmq" | "aws_sqs" | "webhook";
```
alexluong: I'm curious what's the benefit of using the string union here instead of `string`.

I don't quite see the benefit yet and it adds another thing to the list of things to do when a new provider is added.

internal/portal/src/typings/Destination.ts
```diff
@@ -0,0 +1,43 @@
...
+interface Destination {
+  id: string;
+  type: DestinationType;
+  config: Record<string, any>;
+  credentials: Record<string, any>;
+  label: string;
+  description: string;
+  icon: string;
+  instructions: string;
+  target: string;
```
alexluong: These fields look different from the destination model from the BE.

Is the transformation handled in the FE or BE?

alexluong: What's the difference between Destination.target and DestinationTypeReference.target?

internal/portal/src/typings/Destination.ts
alexbouchardd: Yeah fair point

internal/portal/src/typings/Destination.ts
alexbouchardd: This should be the actual value of the "resolved" target

alexbouchardd: My mistake not supposed to be there

PR #161, opened 2024-12-20
feat: Destination config field validation properties

alexluong:
# WIP

- [x] validation properties for config & credentials fields
- [x] server-side validation
- [x] remove previous json-schema setup


internal/portal/src/common/DestinationConfigFields/DestinationConfigFields.scss
```diff
@@ -14,6 +14,13 @@
     }
   }
 
+  input {
+    &:invalid {
+      // TODO: use color variable or update styling
+      background-color: #fee2e2;
+    }
+  }
```
alexluong, line 22: I hard-coded the invalid state like this for now, feel free to make changes as you see fit. I see there's a CSS variable for the background danger but it seems a bit much.

CleanShot 2024-12-20 at 16 23 12
CleanShot 2024-12-20 at 16 25 09

internal/destregistry/metadata/providers/aws_sqs/core.json
```diff
@@ -6,7 +6,8 @@
       "type": "text",
       "label": "Queue URL",
       "description": "The URL of your AWS SQS queue",
-      "required": true
+      "required": true,
+      "pattern": "^(?:https?:\\/\\/)?(?:[\\w\\-]+\\.)+[a-z]{2,}(?::\\d{1,5})?(?:\\/[\\w\\-\\/\\.~:?#\\[\\]@!$&'\\(\\)*+,;=]*)?$"
```
alexluong: For the [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) RegEx here, one thing I noticed was that it was compiled with `v` flag which was quite tricky. Maybe I'm just not a RegEx expert, and unfortunately neither is Claude so we had a bit of a hard time getting something to work.

Figured I'd call this out so you're aware. May worth a notice in the documentation as well in case folks want to override it (or for other devs to take into account when implementing new destination types).

CleanShot 2024-12-20 at 16 29 05

PR #189, opened 2025-01-14
chore: Update examples/docker-compose with latest version & add example YAML config

alexluong:


examples/docker-compose/compose.yml
```diff
@@ -1,39 +1,43 @@
 name: outpost-example
 services:
   api:
-    image: hookdeck/outpost:latest
+    image: hookdeck/outpost:v0.1.0-alpha.5
```
alexluong, line 4: I'm specifying the version here to make sure everyone is using the right version in the example. If the version is `latest`, you will have to manually remove the installed Docker image before it pulls another version when you want to try again. This should fix that issue.

The downside is every time there's a new significant release, we must go and update this value. But that could also be an upside as it forces us to evaluate the impact of new release versions and make sure this example is always compatible with the latest version.

PR #196, opened 2025-01-15
chore: Production-ready router

### LINKED ISSUE #191: Run API service with Gin release mode

alexluong:
resolves #191 

- [x] Run Gin router in release mode by default
- [x] Custom logger middleware

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -35,14 +35,14 @@ func (a *App) Run(ctx context.Context) error {
 
 func run(mainContext context.Context, cfg *config.Config) error {
 	zapConfig := zap.NewProductionConfig()
-	zapConfig.Level = zap.NewAtomicLevelAt(zap.DebugLevel)
+	zapConfig.Level = zap.NewAtomicLevelAt(zap.InfoLevel)
 	zapLogger, err := zapConfig.Build()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


internal/app/app.go
```diff
@@ -34,13 +34,15 @@ func (a *App) Run(ctx context.Context) error {
...
+	zapConfig := zap.NewProductionConfig()
+	zapConfig.Level = zap.NewAtomicLevelAt(zap.DebugLevel)
+	zapLogger, err := zapConfig.Build()
 	if err != nil {
 		return err
 	}
 	defer zapLogger.Sync()
 	logger := otelzap.New(zapLogger,
-		otelzap.WithMinLevel(zap.InfoLevel), // TODO: allow configuration
+		otelzap.WithMinLevel(zap.DebugLevel), // TODO: allow configuration
```
alexbouchardd: Left over todo?

alexluong: It was there from the beginning. It was a bit outside the scope of this PR but we can certainly address it here or in a follow-up PR shortly.

With that said, this change probably shouldn't be included. I log debug something earlier but later remove it, so probably should have reverted this change too. Will update it soon!

alexluong: Quick PR here to address that TODO

https://github.com/hookdeck/outpost/pull/197

Issue #105, opened 2024-11-26
Alerts


alexluong:
## Alerts

Alerts are triggered when an event fails to deliver to a destination. The `ALERT_DEBOUNCING_INTERVAL_MS` variable can configure alerting behaviour. The debouncing interval represents the amount of time that needs to elapse for a new alert to be triggered. Alerts are scoped per destination.

Additionally, destinations can be set to auto-disable after a number of consecutive failures using the `AUTO_DISABLE_FAILURE_COUNT` variable. An alert will also be triggered when the failure count reaches 50%, 70%, 90% and 100% (destination is disabled).

Instead of implementing user-facing alerting, alerts are produced on a callback URL configured through the `ALERT_CALLBACK_URL` variable.

It’s your responsibility to format and deliver the alert to your tenant using your existing notification system.

```json
{
  "topic": "event.failed",
  "disable_treshold": 20,
  "consecutive_failures": 5,
  "destination": DestinationObject,
  "response": {
    "status": "500",
    "data": {
      "some": "value"
    }
  }
}
```

PR #198, opened 2025-01-16
feat: Alert implementation

alexluong:
<details><summary>initial comment</summary>
partial implementation of #105; this PR only implements an `internal/alert` package which is the logic for the alert feature. I'll open another PR shortly where this package is used within the `deliverymq` flow and full e2e test cases.

Note: I just went over the spec and noticed the `ALERT_FAILURE_RATE` feature. Will get around to that next too.

# Alert System Implementation

Implements a Redis-backed alert system for monitoring destination health and auto-disabling failing destinations.

## Core Interfaces

- `AlertMonitor`: Main interface for handling delivery attempts
  - `HandleAttempt(ctx context.Context, attempt DeliveryAttempt) error`

- `AlertStore`: Redis persistence layer for failure counts and alert states
  - `IncrementAndGetAlertState(ctx context.Context, tenantID, destinationID string) (AlertState, error)`
  - `ResetAlertState(ctx context.Context, tenantID, destinationID string) error`
  - `UpdateLastAlert(ctx context.Context, tenantID, destinationID string, t time.Time, level int) error`

- `AlertEvaluator`: Determines when to trigger alerts based on thresholds
  - `GetAlertLevel(failures int64) int`
  - `ShouldAlert(failures int64, lastAlertTime time.Time, lastAlertLevel int) (level int, shouldAlert bool)`

- `AlertNotifier`: Sends HTTP alerts to configured callback URL
  - `Notify(ctx context.Context, alert Alert) error`

## Redis Schema

**Key Structure:**
- `alert:{tenant_id}:{destination_id}:failures` - Counter for consecutive failures
- `alert:{tenant_id}:{destination_id}:last_alert` - Hash storing:
  - `time`: Last alert timestamp
  - `level`: Last alert level (percentage)

## Key Features

- Configurable alert thresholds (e.g. 50%, 70%, 90%, 100%)
- Alert debouncing with configurable interval
- Auto-disable destinations at 100% threshold
- Stateless evaluation with Redis-backed persistence
</details>

### DIFF
--- a/internal/alert/notifier.go
+++ b/internal/alert/notifier.go
@@ -31,6 +31,12 @@ func NotifierWithTimeout(timeout time.Duration) NotifierOption {
 	}
 }
 
+func NotifierWithBearerToken(token string) NotifierOption {
+	return func(n *httpAlertNotifier) {
+		n.bearerToken = token
+	}
+}
+
 // Alert represents an alert that will be sent to the callback URL
 type Alert struct {
 	Topic               string                 `json:"topic"`
@@ -43,6 +49,7 @@ type Alert struct {
 type httpAlertNotifier struct {
 	client      *http.Client
 	callbackURL string
+	bearerToken string
 }
 
 // NewHTTPAlertNotifier creates a new HTTP-based alert notifier
@@ -72,6 +79,9 @@ func (n *httpAlertNotifier) Notify(ctx context.Context, alert Alert) error {
 
 	// Set headers
 	req.Header.Set("Content-Type", "application/json")
+	if n.bearerToken != "" {
+		req.Header.Set("Authorization", "Bearer "+n.bearerToken)
+	}
 
 	// Send request
 	resp, err := n.client.Do(req)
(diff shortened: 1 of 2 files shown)


internal/alert/notifier.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (n *httpAlertNotifier) Notify(ctx context.Context, alert Alert) error {
+	// Marshal alert to JSON
+	body, err := json.Marshal(alert)
+	if err != nil {
+		return fmt.Errorf("failed to marshal alert: %w", err)
+	}
+
+	// Create request
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, n.callbackURL, bytes.NewReader(body))
```
alexbouchardd, line 75: Didn't we say we'd publish to alert to a MQ instead? If we stick to a callback URL we need some auth mecanishm

alexluong: As I recall, we briefly discussed publishing alerts to an MQ but I don't think we made a final call on that. With that said, the way the notifier is implemented we can update the mechanism pretty easily.

I wonder if it makes sense to send the alert to the deliverymq and let the operator register their internal destinations somehow?

Let me know if you'd like us to send the alert to an mq here. I think it's a bit trickier but totally doable too.

Anyway, I did implement the auth callback using API key as the spec, please check it out!

PR #247, opened 2025-02-25
feat: Migration on startup with golang-migrate

alexluong:


leggetter:
@alexluong - if I wanted to QA this, what steps should I follow?

1. Pull branch
2. Build
3. ...?

alexluong:
@leggetter I think the best way to go about it is to run the service with a fresh DB and see if the schema is properly set up.

Another approach is to manually run the migration script with

```
$ go run ./cmd/migrate up
$ go run ./cmd/migrate down
```

But by itself if you don't have your local dev env set up, it's not the easiest to QA. Maybe I can set up something similar to the Docker example so that you can run it locally? It may be easier to QA with #248 too.

PR #248, opened 2025-02-25
feat: Postgres log store implementation + Events API

alexluong:
This PR implements the Postgres logstore + Events API to the spec in Notion.

**Note**: This PR drops support for ClickHouse as it has functionality (mainly the computed `events.status` query) not implemented by ClickHouse yet.


internal/logstore/pglogstore/pglogstore.go
```diff
@@ -11,31 +11,77 @@ import (
...
 
 func (s *logStore) ListEvent(ctx context.Context, req driver.ListEventRequest) ([]*models.Event, string, error) {
 	query := `
-		SELECT id, tenant_id, destination_id, time, topic, eligible_for_retry, data, metadata
-		FROM events 
-		WHERE tenant_id = $1 
-		AND ($2 = '' OR time < $2::timestamptz)
-		ORDER BY time DESC
-		LIMIT $3`
+		SELECT
```
alexluong, line 27: I want to call out this query as it's probably the most complicated part of the implementation for now.

This query is to list events based on a few filters: tenant, destination, status, and cursor-based pagination.

Here's how it's behaving, from my POV:

- This query will query the events first with tenant & destination ID & cursor filter applied.
- If status is provided, for each event it scans, it will evaluate and check whether the event satisfies the condition. That means each evaluated event equates to 1 query in the deliveries table (with index).
- The query is complete when it finds the right number of events based on limit.

To put into number, let’s say a tenant has a 1% failure rate and they query for 100 failed events, that would generates about 10,000 queries in the deliveries table to fully come up with 100 failed events.

From my testing, I tried seeding a DB with a few millions records and this ended up being quite fast because of good indexing. It seems quite a bit better than some alternatives queries like using LEFT JOIN for example.

I also considered using a materialized view for an event_status table or something but not sure how to evaluate the data staleness issue.

Issue #142, opened 2024-12-09
Hookdeck Telemetry


alexluong:
- [ ] Errors & crash report
- [ ] Usage statistics
- [ ] Opt-out

---

Questions:

- How (which tool) should we track error & crash report? Usage statistics?
- What are the key metrics to track?

leggetter:
@alexluong - this is currently labeled as a question. However, I'm assuming there is implementation required here?

alexluong:
@leggetter yes, sorry the "question" label is to mark things that require an answer. Since @alexbouchardd already addressed that, lemme remove the label here. It's still something that needs to be done.

leggetter:
@alexluong disable Sentry for now. Enable with sensible default and configurability in BETA.

alexluong:
implemented #252

PR #252, opened 2025-02-26
feat: Hookdeck Telemetry

alexluong:
implements #142 

## Configuration

Users can set an env `DISABLE_TELEMETRY` to opt-out of this. I wonder if it's better if we support opt-in instead of opt-out mechanism?

## Sentry

We'll init Sentry + use its Gin middleware to capture issues (panic) from the API layer. If there's a 500+ errors, we'll also manually capture them.

As for other flows, we may need to set up custom code to capture each panic. Because we're not using Go's native `recover` feature (yet?), there's no straightforward way to go about this so it's not properly set up for now.

To improve, we should think through Outpost processes and decide which are the important paths that we should recover from potential panic (caused by programming error).

## Hookdeck

One small issue with tracking events with Hookdeck is there's no bulk insert, so we must make a POST request on every event. Since our events shouldn't be too frequent, it may not be a problem for now, but it's something we should consider.

The current implementation uses a buffered channel to separate telemetry handling from the core application logic. From there, we'll send events to Hookdeck in batch, using a few workers to speed up the concurrency. The batch logic could potentially be removed given it doesn't add much value since we can't take advantage of bulk insert anyway.

### Event schema

```
{
  "type": "event_type",
  "time": "TIMESTAMP",
  "payload": { ... }
}
```

### Installation ID

Regarding installation ID, I'm not 100% sure what we should do here. Every time a node restarts, we can generate a new installation ID and use it throughout the lifecycle of that node. Is that okay or should we figure out a way to have a permanent value somehow?


internal/config/config.go
```diff
@@ -149,6 +151,14 @@ func (c *Config) InitDefaults() {
 		ConsecutiveFailureCount: 20,
 		AutoDisableDestination:  true,
 	}
+
+	c.Telemetry = TelemetryConfig{
+		Disabled:          false,
+		BatchSize:         100,
+		BatchInterval:     5,
+		HookdeckSourceURL: "https://hkdk.events/ih7t3hukydzlge",
+		SentryDSN:         "https://examplePublicKey@o0.ingest.sentry.io/0",
```
alexluong, line 162: Can we get the final values that we want to use in production?

internal/config/config.go
```diff
@@ -323,3 +333,28 @@ type AlertConfig struct {
...
+		SentryDSN:         c.SentryDSN,
+	}
+}
+func (c *Config) ToTelemetryApplicationInfo() telemetry.ApplicationInfo {
+	return telemetry.ApplicationInfo{
+		Version:       "TODO",
+		MQ:            "TODO",
+		PortalEnabled: "TODO",
+	}
+}
```
alexluong: Will follow up later with correct information. It's a matter of parsing the config into the data we want.

If you have any further suggestion for what information we care about, please mention them here.

internal/config/config.go
```diff
@@ -323,3 +333,28 @@ type AlertConfig struct {
...
 	return c.configPath
 }
+
+type TelemetryConfig struct {
+	Disabled          bool   `yaml:"disabled" env:"DISABLE_TELEMETRY"`
+	BatchSize         int    `yaml:"batch_size" env:"TELEMETRY_BATCH_SIZE"`
+	BatchInterval     int    `yaml:"batch_interval" env:"TELEMETRY_BATCH_INTERVAL"`
+	HookdeckSourceURL string `yaml:"hookdeck_source_url" env:"TELEMETRY_HOOKDECK_SOURCE_URL"`
+	SentryDSN         string `yaml:"sentry_dsn" env:"TELEMETRY_SENTRY_DSN"`
+}
```
alexluong, line 345: During local development, you can customize these values with whatever you prefer. For example, I have this in my YAML config to send to my Sentry and Hookdeck

```
telemetry:
  disabled: false
  batch_size: 100
  batch_interval: 5
  hookdeck_source_url: "..."
  sentry_dsn: "..."
```

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Wait for all workers
+	wg.Wait()
+}
+
+func (t *telemetryImpl) Init(ctx context.Context) {
+	if err := sentry.Init(sentry.ClientOptions{
+		Dsn:              t.config.SentryDSN,
+		EnableTracing:    true,
+		TracesSampleRate: 1, // FIXME
```
alexluong: Which value should we use here?

leggetter:
> Installation ID
> Regarding installation ID, I'm not 100% sure what we should do here. Every time a node restarts, we can generate a new installation ID and use it throughout the lifecycle of that node. Is that okay or should we figure out a way to have a permanent value somehow?

@alexluong this should be a per-installation unique ID. So, generated on the first run (install) and then persisted between start/stop events.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,229 @@
...
+		"mq":             a.MQ,
+		"portal_enabled": a.PortalEnabled,
+	}
+}
+
+type telemetryEvent struct {
+	EventType string                 `json:"type"`
+	Payload   map[string]interface{} `json:"payload"`
+	Time      time.Time              `json:"time"`
+}
```
leggetter: I think the Installation ID should be at this top level along with `type` and `payload`.

alexluong:
> @alexluong this should be a per-installation unique ID. So, generated on the first run (install) and then persisted between start/stop events.

@leggetter how should we go about persisting this ID?

internal/telemetry/telemetry.go
leggetter: `time` -> `timestamp`. Assuming this is an ISO 8601 timestamp?

internal/config/config.go
leggetter: For Hookdeck: https://hkdk.events/yhk665ljz3rn6l

We'll need @alexbouchardd for the Sentry URL

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,229 @@
...
+			}
+			if len(batch) > 0 {
+				t.sendBatch(batch)
+			}
+			return
+		}
+	}
+}
+
+func (t *telemetryImpl) sendBatch(batch []telemetryEvent) {
```
leggetter, line 96: Are we batching to send and then sending periodically? We're not batching the events on the outbound HTTP/API request, right?

I want to send the telemetry events to PostHog, so I'm assuming each HTTP request to Hookdeck is a single event. It won't work otherwise.

```js
addHandler('transform', (request, context) => {

  // PostHog event format
  // {
  //   "api_key": "PH_API_KEY",
  //   "event": "event name",
  //   "distinct_id": "user distinct id",
  //   "properties": {
  //     "account_type": "pro"
  //   },
  //   "timestamp": "[optional timestamp in ISO 8601 format]"
  // }

  // Outpost Telemetry format
  // type telemetryEvent struct {
  //   EventType string                 `json:"type"`
  //   Payload   map[string]interface{} `json:"payload"`
  //   Time      time.Time              `json:"time"`
  // }

  const captureEvent = {
    "api_key": process.env.POSTHOG_API_KEY,
    "event": request.body.type,
    "distinct_id": request.body.installation_id,
    "properties": request.body.payload,
    "timestamp": request.body.time,
  }

  request.body = captureEvent;

  return request;
});
```

internal/telemetry/telemetry.go
leggetter: nit: I feel like `payload` should just be `data`.

leggetter:
> > @alexluong this should be a per-installation unique ID. So, generated on the first run (install) and then persisted between start/stop events.
> 
> @leggetter how should we go about persisting this ID?

@alexluong - what options do we have? DB or a type of `.outpost` file on disk?

alexluong:
I don't think storing to disk would work since it's ephemeral. Maybe storing the installation in Redis could be an option?

leggetter:
> I don't think storing to disk would work since it's ephemeral. Maybe storing the installation in Redis could be an option?

@alexluong as long as Redis is set up to persist the data 👍

internal/telemetry/telemetry.go
alexluong: > Are we batching to send and then sending periodically? We're not batching the events on the outbound HTTP/API request, right?

I'm not 100% sure I understand your questions here. We are batching and sending the events periodically, but currently each batch is just a loop and then HTTP POST individual events.

> so I'm assuming each HTTP request to Hookdeck is a single event

Yes, this is how it works currently. Seems like the current setup should work for you.

internal/telemetry/telemetry.go
alexluong: Updated in the latest version, can you go over them and resolve this if the change looks good?

internal/config/config.go
alexluong: @leggetter any suggestion here?

internal/app/installation.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	// TODO: consider using WATCH to avoid race condition
+	// There's a potential race condition when multiple Outpost instances are started at the same time.
+	// However, given this is for telemetry purposes, and it will be a temporary issue, we can ignore it for now.
+	installationID, err := redisClient.HGet(ctx, outpostrcKey, installationKey).Result()
+	if err != nil {
+		if err == redis.Nil {
+			installationID = uuid.New().String()
+			if err = redisClient.HSet(ctx, outpostrcKey, installationKey, installationID).Err(); err != nil {
```
alexluong, line 33: I'm thinking of adding a hash to Redis, something like `outpostrc`, and for now it only has `installation_id` as its only value.

If the user disables telemetry, then we won't add this hash.

What do you think about this approach?

internal/config/config.go
leggetter: Nothing additional for now.

internal/telemetry/telemetry.go
leggetter: @alexluong - I'm not sure what you're looking for. Can you provide more info/more context?

Issue #257, opened 2025-03-10
Improve Kubernetes example


leggetter:
- [ ] Complete Chart.yaml
- [ ] Add NOTES.txt
- [ ] Add values.schema.json
- [ ] Evaluate further configuration options

PR #258, opened 2025-03-10
k8s quickstart

leggetter:
Changes:

- Moves the kubernetes example from `deployments` into `examples`
- Move the README into `docs/quickstarts/kubernetes.mdx`
- Updates the Docker quickstart to follow the outline in the Kubernetes one
- Moved the TODO items in the k8s README into https://github.com/hookdeck/outpost/issues/257


docs/pages/quickstarts/docker.mdx
```diff
@@ -0,0 +1,105 @@
+---
+title: "Outpost Quickstart: Docker with Rabbit MQ or AWS SQS via LocalStack"
```
alexluong: I think it should just be `RabbitMQ`, I don't see folks splitting them up.

alexluong (commented):
I think this approach is fine. The only issue tho is the whole `charts/outpost` thingy. It's a dedicated Helm chart that will be a releasable artifact, not just for an example.

Ideally, in an example, the workflow should be something along this line:

```
helm repo add hookdeck <hookdeck_helm_charts_repo_url>
helm install outpost hookdeck/outpost ...other_flags
```

docs/pages/quickstarts/docker.mdx
leggetter: This single quickstart covers using two different queuing options.

1. RabbitMQ
2. SQS via LocalStack

Are you suggesting we split them up?

leggetter:
> I think this approach is fine. The only issue tho is the whole `charts/outpost` thingy. It's a dedicated Helm chart that will be a releasable artifact, not just for an example.
> 
> Ideally, in an example, the workflow should be something along this line:
> 
> ```
> helm repo add hookdeck <hookdeck_helm_charts_repo_url>
> helm install outpost hookdeck/outpost ...other_flags
> ```

Ok, captured in https://github.com/hookdeck/outpost/issues/257

docs/pages/quickstarts/docker.mdx
alexluong: No, I meant `RabbitMQ` instead of `Rabbit MQ`

PR #281, opened 2025-03-17
fix: Destination config input supports parsing into map string

alexluong:
Destination `config` and `credentials` are both a string map. To support boolean better, we should automatically parse `true -> "true"`. This was already supported for credentials but not for config yet. This PR is to support both.

Example:

Before:

```
POST /:tenantID/destinations
{
  ...
  "config": {
    "bool": true,
    "obj" : { "hello": "world" }
  }
}

400 error
{
  "message": "invalid JSON"
}
```

With this PR, the destination above will be parsed into

```
{
  ...
  "config": {
    "bool": "true",
    "obj" : "{\"hello\":\"world\"}"
  }
}
```

### DIFF
--- a/internal/services/api/destination_handlers.go
+++ b/internal/services/api/destination_handlers.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"errors"
-	"log"
 	"net/http"
 	"strings"
 	"time"
@@ -73,7 +72,6 @@ func (h *DestinationHandlers) List(c *gin.Context) {
 func (h *DestinationHandlers) Create(c *gin.Context) {
 	var input CreateDestinationRequest
 	if err := c.ShouldBindJSON(&input); err != nil {
-		log.Println(err)
 		AbortWithValidationError(c, err)
 		return
 	}
--- a/internal/services/api/destination_handlers.go
+++ b/internal/services/api/destination_handlers.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"errors"
+	"log"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/services/api/destination_handlers.go
```diff
@@ -72,6 +73,7 @@ func (h *DestinationHandlers) List(c *gin.Context) {
 func (h *DestinationHandlers) Create(c *gin.Context) {
 	var input CreateDestinationRequest
 	if err := c.ShouldBindJSON(&input); err != nil {
+		log.Println(err)
```
leggetter: Checking if this is dev-debug that's crept in? Do we tend to log the errors within the error if statement? Is that a pattern we follow?

alexluong: oops no this is my bad, if we do log we would use a proper logger instead of this

fixed

leggetter:
@alexluong, this seems to have a knock-on effect to the DX, API(?), docs, and examples. Is that correct?

alexluong:
Can you clarify, I'm not sure I understand what you mean?

leggetter:
> Can you clarify, I'm not sure I understand what you mean?

I'm trying to understand what changes this has for the person managing their installation of Outpost. It's unclear to me if that person now needs to set config as a JSON string rather than simply JSON. Or if this is only an internal change that doesn't leak out to the DX. If it does leak out to the DX I'd like to see how that changes what we have in the docs and examples.

leggetter:
@alexluong feel free to merge this one

PR #285, opened 2025-03-19
chore: use event.topic as routing_key for rabbitmq

leggetter:
- Removes `routing_key` from config
- Uses `event.topic` with `PublishWithContext` as routing key
- Adds mock testing for `PublishWithContext` and had to add a test method and export two interfaces to enable it
- Renamed the existing publish test to be suffixed with `_integration_test` within it, which has confused the diff

`make test/unit` passes. However, `make test` fails. However, this also happens on the `main` branch.


alexluong:
Hi @leggetter, I think we should probably have connected a bit more on the test suite. `make test` should pass but it's not super simple since we have quite a bit of tests and it can be a bit flaky, especially if you don't have the right setup.

Before we get into that, let's elaborate on the destination test suite. In our current setup, we have a generic publisher test suite that every destination must pass. That was in the `destrabbitmq_publish_test.go` file that I think you removed and set up your own mock test.

The idea of the publisher suite is that each destination must provide its own consumer, and the publisher will run that consumer and assert on each message. The suite have various publishing behavior so that it can test it more comprehensively and uniformly for each destination.

---

Back to the e2e test suite, we do have some dev documentation like so

CleanShot 2025-03-21 at 14 25 26

Happy to chat more if that helps. In the meantime, I can look more into this and get it working.

leggetter:
@alexluong 

> I think you removed and set up your own mock test.

No, just renamed it.

internal/destregistry/providers/destrabbitmq/destrabbitmq_publish_test.go
```diff
@@ -1,188 +1,138 @@
...
-		s.consumer.Close()
-	}
-}
-
-func TestRabbitMQPublish(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping integration test in short mode")
-	}
-	suite.Run(t, new(RabbitMQPublishSuite))
-}
```
alexluong: This is the code that initiate running the integration test for this destination as I described above.

@leggetter

leggetter: Yeah, it's moved into another file.

leggetter: https://github.com/hookdeck/outpost/pull/285/files#diff-7cdc978e39f75b93a912d427fb1862ed46b3913ab3fb5d339f82b5aff6199d8fR170

alexluong: @leggetter I see, sorry my bad I somehow missed that.

Anyway, if you don't mind, I removed the unit test you set up in favor of the original integration test for consistency purposes. I updated that test to work with the new routing key convention, so please feel free to review if you're interested in how it works.

Do let me know if you feel the unit test is helpful moving forward, we can certainly revert and include that if that helps your workflow. I'm also around if you'd like to connect and discuss the workflow further and how to run the integration test suite.

PR #297, opened 2025-03-26
fix: Send system metadata to destrabbitmq & destawssqs

alexluong:


internal/destregistry/basepublisher.go
```diff
@@ -30,3 +34,19 @@ func (p *BasePublisher) StartClose() {
 	p.closed.Store(true)
 	p.active.Wait()
 }
+
+func (p *BasePublisher) MakeMetadata(event *models.Event, timestamp time.Time) map[string]string {
```
alexluong, line 38: I added this helper to use in each destination. However, I'm not using it in `destwebhook` intentionally because it has configs related to disabling default headers. If using this helper, it won't be able to tell whether a key-value is default (our system metadata) or user-provided event metadata.

PR #315, opened 2025-04-04
feat: OpenAPI spec

leggetter:
TODO: 

- [x] Identify payloads that change based on `type`
- [x] Add details for payloads based on `type`


leggetter:
@alexluong whilst creating the OpenAPI spec from the router, it seems that a lot of methods are mapped to the root of the API as well as `/:tenantId`. Which of these methods should we include in the OpenAPI spec and which should we hide as convenience methods? Or should we list everything for completeness?

See https://outpost-docs-git-feat-api-spec-hookdeck.vercel.app/docs/docs/api/health?_vercel_share=s3MpRPmywG67hbsA5fFUXoc59nGWElrx

CleanShot 2025-04-04 at 20 11 58@2x

alexluong:
Hmm it's a bit tricky to say. The ones which are mapped to root & tenants are root, with tenant being a helper endpoint for the portal.
I think we can just document the root ones since you don't really need the tenant ones. If you'd like, you can also mention it in the description or something if that helps?

Issue #305, opened 2025-04-01
Destination - Amazon Kinesis


leggetter:
Publish events to Amazon Kinesis https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html#API_PutRecord_Examples

TODO:

- [ ] QA
- [ ] Documentation

alexluong:
I explored AWS Kinesis a bit and came up with this design document here. Can you take a look? The main 3 things to note here

- Kinesis doesn't support metadata natively
- How to decide `partition_key`
- Batch processing support (probably in a later iteration)

<details>
<summary>RFC</summary>

# AWS Kinesis Destination Design

## Overview

AWS Kinesis is a managed service for real-time data streaming that enables:
- Ingestion and processing of streaming data at scale
- Temporary storage of data (default 24h, configurable up to 365 days)
- Partitioned data flow through shards (capacity units)

Kinesis requires:
- Stream name
- AWS region
- Partition key strategy
- AWS credentials

## Implementation Approach

### Core Operation: PutRecord

For the initial implementation, the single-record `PutRecord` API will be used which:
- Sends one record at a time
- Requires stream name, partition key, and data payload
- Aligns with Outpost's current per-event publishing model

```go
// AWS SDK PutRecord operation
PutRecordInput {
    StreamName: "my-stream",
    PartitionKey: "computed-key",
    Data: []byte(eventJSON)
}
```

Future optimization could include `PutRecords` for batching multiple events in a single call.

## Metadata & Data Payload

Kinesis records are limited to 1MB and contain:
- Data payload (event data as JSON)
- Partition key (determines shard assignment)
- Sequence number (assigned by Kinesis)
- Timestamp (assigned by Kinesis)

Outpost event metadata will be included in the data payload itself.

## Credentials

Required AWS credentials follow standard pattern:

```json
{
  "key": "AWS_ACCESS_KEY_ID",
  "secret": "AWS_SECRET_ACCESS_KEY",
  "session": "AWS_SESSION_TOKEN" // Optional
}
```

- All fields should be treated as sensitive
- Session token is optional (used for temporary credentials)

## Configuration

Essential configuration parameters:

```json
{
  "stream_name": "my-kinesis-stream", // Required
  "region": "us-east-1", // Required
  "partition_key_template": "{{event.topic}}", // Optional, defaults to event.topic
  "endpoint": "https://kinesis.us-east-1.amazonaws.com" // Optional, for custom endpoints
}
```

[...]

PR #316, opened 2025-04-07
feat: Support AWS Kinesis destination

alexluong:
This implements the RFC from #305. A few note-worthy things to call out here:

## 2. `config.endpoint` UI

This should probably be hidden. However, it would make local dev testing difficult, especially when using the UI to test locally. With SQS, we can derive the endpoint from the `queue_url` field and detect whether it's production AWS or local LocalStack. We can't do the same with AWS Kinesis.

CleanShot 2025-04-08 at 01 25 31

Options:
1. Hide it and assume production AWS 
2. Show it and provide default to AWS

## 3. Partition key template payload

This PR implements the partition templating with JMESPath as discussed in the RFC.

One thing to note is that I ended up going with the message payload as the input for the template. What I mean is that when producing a Kinesis record, we need 3 info: `stream_name`, `partition_key`, and `data` (bytes version of a Go map); whatever the data content is, that map is the payload for the partition key template.

For example, with an event like so:

```
{
    "id": "evt_1",
    "tenant_id": "te_1",
    "topic": "user.created",
    "eligible_for_retry": false,
    "metadata": {
        "mymeta": "metadata"
    },
    "data": {
        "data": "data"
    }
}
```

If `metadataInPayload: true`, the resulting payload would be:

```
{
  "metadata": {
    "mymeta": "metadata",
    "event-id": "evt_1",
    "topic": "user.created",
    "timestamp": "123456"
  },
  "data": {
    "mydata": "data"
  }
}
```

So, you can do something like `metadata.mymeta` or `metadata.topic` as the partition key tempalte.

If `metadataInPayload: false`, the resulting payload would be

```
{
  "mydata": "data"
}
```

In that case, you can do `mydata` for the template. Or an empty string, in which case we'll use the event ID as the partition key. One note is that you will NOT be able to use topic as the partition key in this case, unless topic is part of the data itself.

### Reasoning

The reason I went with this approach was around the end-user perspective. They won't know what an event is, and the only thing they know is the payload they'll receive. Therefore, I think this makes a bit more sense than the incoming event payload from the producer/operator side.

However, this still requires the user to know beforehand what the expected data shape is. It's a bit more complicated than webhook, for example, where you can just provide a URL and observe the incoming requests. Maybe it's a necessary consideration, given the technology.

## 4. Delivery details

CleanShot 2025-04-08 at 01 31 49

This is how the UI looks right now.

**NOTE**: The "message" UI is incorrect now. Currently, the message UI is fake, deriving from the event. For Kinesis, it doesn't account for the fact that the user can opt-out of metadata, and therefore the message can be different. We probably need to add a task to persist message data & display properly.

---

Dev notes:

- When testing locally, I added a script to help provision a stream and log incoming events: `go run ./cmd/destinations/awskinesis`. Make sure you run `make up/deps` beforehand. There're new changes to the Docker Compose, so if you had it running previously, also make sure to restart the containers.

---

TODO:

- [ ] QA & go over DX/UX
- [ ] Instructions & re-write config descriptions / helper texts

---

Next step:

- [ ] Consider `PutRecords` to send events in batch
- [ ] Consider ways to improve UX around templating


leggetter:
> 1. config.metadata_in_payload

I don't think this should be exposed to the end-user. I think it's operator-only config as the operator should determine what's in the event payload and not the end-user.

> 2. config.endpoint UI

Show it for now. 

> 3. Partition key template payload

I don't quite understand this. But I'm assuming my comment on 1. would have an impact.

May be something worth discussing live and then capturing the outcome in GitHub.

> 4. Delivery details

Gotcha. End logic may be impacted by 1. and 3.

PR #327, opened 2025-04-11
feat: Support GCP PubSub for publishmq & internalmq

### LINKED ISSUE #138: InternalMQ (Delivery & Log) - GCP Pub/Sub
Built on top of https://github.com/hookdeck/outpost/issues/325

alexluong:
resolves #138 #325

This PR adds support for Google Cloud Pub/Sub as a message queue provider for both internal MQs and publish MQ.

## Config

Config for internalmq along with the default values:

```
GCP_PUBSUB_PROJECT=""
GCP_PUBSUB_SERVICE_ACCOUNT_CREDENTIALS=""
GCP_PUBSUB_DELIVERY_TOPIC="outpost-delivery"
GCP_PUBSUB_DELIVERY_SUBSCRIPTION="outpost-delivery-sub"
GCP_PUBSUB_LOG_TOPIC="outpost-log"
GCP_PUBSUB_LOG_SUBSCRIPTION="outpost-log-sub"
```

The only required config is `GCP_PUBSUB_PROJECT`.

For publishmq, there's no default, so if you want to use it, you must set all config:

```
PUBLISH_GCP_PUBSUB_PROJECT=""
PUBLISH_GCP_PUBSUB_SERVICE_ACCOUNT_CREDENTIALS=""
PUBLISH_GCP_PUBSUB_TOPIC=""
PUBLISH_GCP_PUBSUB_SUBSCRIPTION=""
```

## Local emulator

This PR added Google Cloud emulator to the dev Docker Compose (outpost-deps). To test locally, make sure you set the env `PUBSUB_EMULATOR_HOST="gcp:8085"`.

## Cloud env

There are 2 ways to authenticate with GCP Cloud. You can either rely on the automatic credentials mechanism, using `GOOGLE_APPLICATION_CREDENTIALS` env pointing to the credentials file, or use some sort of IAM / Workload Identity in production. The last resort is to use the "service account credentials" variable (`GCP_PUBSUB_SERVICE_ACCOUNT_CREDENTIALS` for internalmq or `PUBLISH_GCP_PUBSUB_SERVICE_ACCOUNT_CREDENTIALS` for publishmq). This env expects the escaped JSON string which you can get by doing something like `jq tostring ./path/to/keys.json`.

## Publishmq local helper

We have `cmd/publish` server which is a helper dev server so you can test the publishmq locally easily. Once the server is running `go run ./cmd/publish` on port 5555, you can use curl or Postman to interact with it like so:

```
# provision the topic + subscription
curl --location --request POST 'localhost:5555/declare?method=gcp_pubsub'

# send an event to publishmq
curl --location 'localhost:5555/publish?method=gcp_pubsub' \
--header 'Content-Type: application/json' \
--data '{
    "tenant_id": "12345",
    "topic": "user.created",
    "eligible_for_retry": true,
    "metadata": {
        "meta": "data"
    },
    "data": {
        "user_id": "userid"
    }
}'
```

## PublishMQ

For publishmq, the user MUST create and manage the topic & subscription themselves. All we do is subscribing to it and processing the incoming messages.

NOTE: for publishmq with GCP, technically we only care about the Subscription. In theory, we don't care about the Topic at all, but because of the way the mq logic is implemented, we do require topic to be correct. In the future, we can consider refactor the mq (as mentioned in another convo) so that we don't need to care about the publish topic anymore.


internal/mqinfra/gcp.go
alexbouchardd, line 1: nitpick: Shouldn't this file be called `gcp_pubsub.go`?

alexluong: @alexbouchardd good call, I updated that along with a few other files to specify AWS SQS and GCP PubSub instead of just AWS / GCP.

alexbouchardd (approved):

internal/mqinfra/gcp.go
alexbouchardd: Great!

leggetter:
@alexluong - tested as internal MQ and as publish queue via GCP. Good to merge 👍

PR #334, opened 2025-04-17
Feat/portal manual retry

leggetter:
@alexluong - adding manual retry to the portal in this branch.

Should we update the API behavior in this branch, too?

Clipboard-20250417-170707-454


alexluong:
> Should we update the API behavior in this branch, too?

Up to you, I think maybe it's easier to split into multiple PRs so we can merge things faster but it's totally fine if you'd like to includer that here.

Issue #317, opened 2025-04-07
Portal destination config checkbox bug


alexluong:
The portal destination config & credentials form is a configurable form based on the destination metadata. The checkbox (or boolean) field has a few bugs or issues related to the way *uncheck* HTML checkbox works.

From [HTML input checkbox MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox):

> Note: If a checkbox is unchecked when its form is submitted, neither the name nor the value is submitted to the server. There is no HTML-only method of representing a checkbox's unchecked state (e.g., value=unchecked). If you wanted to submit a default value for the checkbox when it is unchecked, you could include JavaScript to create a [<input type="hidden">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden) within the form with a value indicating an unchecked state.

and this is how we're constructing the values in this form:

```tsx
<form
  onSubmit={(e) => {
    e.preventDefault();
    const formData = new FormData(e.target as HTMLFormElement);
    const values = Object.fromEntries(formData.entries());
    // ...
  }}
/>
```

Therefore, when the field is unchecked, its value is not passed to the payload. For example, let's say for RabbitMQ, we have `config.tls` as a checkbox / boolean value. When checked, the config should have

```
{
  "tls": "on"
}
```

When unchecked, `tls` is not part of config.

This leads to 2 problems

1. When creating new destination, we cannot support a bool whose default is false. For example, in the AWS Kinesis destination, we want to support `config.metadataInPayload` default to `false`. However, it's simply not possible.

2. The way our update logic works, we merge the existing config with the new config input to create a new updated config. Therefore, to change something to from `true` -> `false`, it must be passed explicitly to the payload. With the limitation described above, this is impossible.

alexluong:
To be completely honest, I'm not 100% sure what's the best way to solve this yet. Will noodle on it some more. Just want to bring this up as an issue. Any ideas are super welcome!

PR #335, opened 2025-04-22
fix(portal): ensure checkbox values are present as true/false. Fix handling of true/false value in DestinationConfigFields.

leggetter:
Ensures that checkbox values have an entry within FormData when checked or unchecked by using a utility function.

Also, use a utility function to determine whether a value represents checked.

Fixes https://github.com/hookdeck/outpost/issues/317


internal/portal/src/utils/formHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+ * isCheckedValue("on");
+ * 
+ * @example
+ * // Returns false
+ * isCheckedValue("false");
+ * isCheckedValue("");
+ */
+export function isCheckedValue(value: string | undefined): boolean {
+  return value === "true" || value === "on";
+}
```
alexluong: I think there should be an extra line here if we use Prettier here. Is that something we care about in the codebase?

alexluong (commented):
Looks great, confirmed this fixes the bug where we were unable to change from true -> false.

Should we consider making something like rabbitmq config.tls default to true, given that should be the expectation for production system? Beside that field, I don't think there's any other checkbox field for now.

alexluong (approved):

leggetter:
> Should we consider making something like rabbitmq config.tls default to true, given that should be the expectation for production system? Beside that field, I don't think there's any other checkbox field for now.

In internal/destregistry/metadata/providers/rabbitmq/metadata.json `tls` does default to `on`.

I considered making all checkboxes default to checked, but that isn't the standard checkbox behavior.

Issue #136, opened 2024-12-09
Destination Hookdeck Event Gateway


alexluong:

alvarp:
Hey! Regarding the managed sources, Outpost should redirect to `https://dashboard.hookdeck.com/connect` and this will allow users to create a managed source with a token directly.

I'm not fully sure about how the token should be used in the auth or if this is already implemented on the Outpost side, but I suspect that, similar to the dashboard, it is already done.

Demo:

https://github.com/user-attachments/assets/38d8f7dc-9a68-44f7-998d-863790587020

PR #336, opened 2025-04-23
feat: desthookdeck

alexluong:
This PR implements #136. Integration summary:

- The portal will use metadata `setup_link` to determine how to render the config & credentials forms. Please go over the screenshots to see how this feels. I think we may need to update something here, particularly the update flow.
- The destination will store the token at `credentials.token`. Before creating the destination, Outpost will verify that the token is valid and store the name at `config.name` (readonly). This name is refreshed whenever the token is updated.

## Questions

1: From @alexbouchardd 

> In the /connect link we should be appending a query of ?provider=${organization_name} the ORGANIZATION_NAME is a outpost variable. The connect UI handles the query param to display the name of the provider.

I'm not 100% sure how we should go about this since the setup link href is from the metadata.json, so we can't append the organization name into it. Should we have a special check somewhere to apply this?

For example, the desthookdeck can override the base provider "Metadata" method to apply the query param to the setup link? I'm not sure if that's a bit too edge case and unexpected for end-users, although maybe they may not care?

### Destination create / update

CleanShot 2025-04-23 at 13 34 18
CleanShot 2025-04-23 at 13 34 50
CleanShot 2025-04-23 at 13 35 04
CleanShot 2025-04-23 at 13 35 25

### Destination target (copyable value)
CleanShot 2025-04-23 at 13 37 25

This is how we're showing this section currently. If the copy button is clicked, the copied value is "Hookdeck source: src_GcxGhypwnBIX". I wonder if it's better to use the full Hookdeck source URL or just the source ID instead?

### Delivery response

CleanShot 2025-04-23 at 13 39 13

Any thoughts on what we should store for the delivery response? We're treating it as an HTTP request currently but maybe a custom Hookdeck-specific schema is better?


leggetter:
@alexluong - I also think that we'll want to be able to track that the connect came from an outpost deployment. Of course, only Outpost uses this at present. But I see there are plans to open this connect flow up to other providers in the future.

leggetter:
@alexbouchardd - The connect flow should support choosing a project for the source to be created in.

leggetter:
@alexbouchardd - why did we choose a custom auth flow over something like OAuth2? The need to copy the source token feels clunky.

alexbouchardd:
> why did we choose a custom auth flow over OAuth2? The need to copy the source token feels clunky. I had an expectation for the flow to auto-populate the information in Outpost.

@leggetter There's 3 main reasons for this:

1. Oupost deployments are anonymous, and we can't provide Oauth credentials without having them "register" in some way with Hookdeck
2. Outpost deployments don't have to be publicly exposed so there's no viable callback URL post oauth flow
3. We could fix #2 by doing a callback to the UI and have the UI handle it, but then that puts an additional burden on developers building their own UI only to support Hookdeck. We didn't want Hookdeck to add a development burden. 

> The connect flow should support choosing an org and project for the source to be created in. It can default to the user's current values, but should allow for them to be changed.

Yeah, fair point, we'll add it. It's using the same UI as the CLI flow, and I've had the same thought there. I'll change the layout to support changing org.

---

@alexluong Can we control the display order of the destinations and the ordering returned by the schema endpoint? Order should be Webhooks -> Hookdeck -> SQS -> Kinesis -> RabbitMQ

--- 

@alexluong The intention was to display the source name, not id. +1 for a link to the Hookdeck dashboard.

leggetter:
https://github.com/hookdeck/outpost/pull/336/commits/e10a2df198bbbc9080afe15af4d98d9d3cc9f26d shows the *Configuration Guide* button all the time. i.e. it's no longer hidden if the setup link is present.

alexluong:
Added a few commits to address the feedbacks here

> Can we control the display order of the destinations and the ordering returned by the schema endpoint? Order should be Webhooks -> Hookdeck -> SQS -> Kinesis -> RabbitMQ

> The intention was to display the source name, not id. +1 for a link to the Hookdeck dashboard.

Here's how the UI looks. When clicking on the ID, it will open the Hookdeck dashboard URL.
CleanShot 2025-04-24 at 17 05 49@2x

cc @leggetter @alexbouchardd 

---

## Breaking change

I updated the Hookdeck config schema a bit to go from `config.name` to `config.source_name` and `config.source_id`. I think it's a bit clearer that way, in case we have a "destination name" in the future, what do you think?

alexbouchardd:
Looks good 🔥

Issue #311, opened 2025-04-03
Local benchmark setup


leggetter:
Scenario: 1000 webhooks per second locally for 60 seconds

---

Current config setup:

- Number of tenants, each has one destination
- Number of events

PR #342, opened 2025-04-29
chore: Local loadtest with k6

alexluong:
implements #311


loadtest/docs/README.md
```diff
@@ -0,0 +1,48 @@
+# Outpost Loadtesting Documentation
```
alexluong, line 1: Loadtest instructions here

loadtest/docs/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+- **Prometheus & Grafana**: Monitoring tools to observe resource utilization during tests, helping identify bottlenecks and performance issues.
+
+- **Mock Webhook Destination**: A simulated event sink used to verify event delivery and measure end-to-end latency.
+
+- **k6 Testing Scripts**: Two main testing flows:
+  - **Events Throughput**: Publishes events to the Outpost service to test throughput capacity.
+  - **Events Verify**: Checks the mock webhook destination to verify successful delivery of events.
+  
+  These scripts are coordinated using Redis to maintain test state and correlate published events with verifications.
```
leggetter, line 35: Is there something that outputs:

1. Throughput in messages/second: totalMessages / toSeconds(lastMessageTimestamp - firstMessageTimetamp)
2. Latency:
    - P50 (Median): Half of the messages were faster than this value.
    - P95: 95% of messages were faster; 5% were slower.
    - P99: Useful to catch tail latency (rare but potentially critical delays).
    
?

alexluong: Yes, the events-throughput one is to simulate load. You can configure the rate (rps, request = publish) using the config. It will output the number of requests sent along with avg rps.

events-verify will randomly select a few (the count configurable via MAX_ITERATIONS env) to verify that it's delivered. This script will output the event latency metrics (with max, avg, and p90, p95). Keep in mind this metrics is the end-to-end latency, with publish timestamp - received timestamp. We can potentially get the latency from when outpost receives the event til it's being delivered too, but it's not implemented yet.

PR #344, opened 2025-04-30
chore(docs): add hookdeck and kinesis types to OpenAPI spec

leggetter:
Implements #343

Note: should we remove the `config` from Hookdeck?

### DIFF
--- a/docs/apis/openapi.yaml
+++ b/docs/apis/openapi.yaml
@@ -66,14 +66,6 @@ components:
         success:
           type: boolean
           example: true
-    DestinationOAuthRedirect:
-      type: object
-      properties:
-        redirect_url:
-          type: string
-          format: url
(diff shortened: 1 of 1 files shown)


alexluong:
> Note: should we remove the config from Hookdeck?

Currently, `config` is a required field for every destination, even if there's no configuration. Should we make it optional then?

leggetter:
> > Note: should we remove the config from Hookdeck?
> 
> Currently, `config` is a required field for every destination, even if there's no configuration. Should we make it optional then?

Is there a way to make it optional for Hookdeck and other destinations where there isn't any config to add?

Not a priority.

alexluong:
> Is there a way to make it optional for Hookdeck and other destinations where there isn't any config to add?

Yes, PR #348 here, let me know if that works for you.

PR #347, opened 2025-05-02
feat: Support custom user-agent for HTTP-based destinations

### LINKED ISSUE #262: Override default webhook user-agent header and allow it to be configured
The current `user-agent` is `Go-http-client/2.0`. We should override with `outpost/{VERSION}`.

Then, we should allow that to be configurable via ENV and YAML

alexluong:
resolves #262 

The PR allows users to configure their user-agent headers for HTTP-based destinations.

The user agent template is `PRODUCT/VERSION`. Version is the Outpost service version. The PRODUCT is configurable, default to `Outpost`. The user can either use env `DESTINATIONS_USER_AGENT_PRODUCT=Acme` or the YAML config

```
destinations:
  user_agent_product: "Acme"
```

The thought process behind "user_agent_product" is that it's the [product](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent) part of the user agent, using terminology from the web platform (MDN). The prefix `destinations` is consistent with destinations-related configuration.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -54,6 +54,8 @@ type Config struct {
 	// Application
 	AESEncryptionSecret string   `yaml:"aes_encryption_secret" env:"AES_ENCRYPTION_SECRET"`
 	Topics              []string `yaml:"topics" env:"TOPICS" envSeparator:","`
+	OrganizationName    string   `yaml:"organization_name" env:"ORGANIZATION_NAME"`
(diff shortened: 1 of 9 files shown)


alexbouchardd:
@alexluong We need to support overriding the user agent altogether for backward compatibility. I think we should use `ORGANIZATION_NAME` for `product` and the outpost `version` for `version` UNLESSS the `HTTP_USER_AGENT` is set in which case we use it as is (version included)

alexbouchardd:
👍

PR #376, opened 2025-05-09
chore: AWS deployment for benchmarking

alexluong:


loadtest/src/tests/events-verify.ts
```diff
@@ -179,32 +215,170 @@ export default async function () {
...
+          //     receivedTimestampMs && deliveryStartTimestampMs
+          //       ? receivedTimestampMs - deliveryStartTimestampMs
+          //       : "N/A"
+          //   }
+          //   Total calculated: ${totalCalculatedLatency}
+          //   Diff from E2E: ${
+          //     receivedTimestampMs && sentTimestamp
+          //       ? receivedTimestampMs - sentTimestamp - totalCalculatedLatency
+          //       : "N/A"
+          //   }`);
```
leggetter, line 374: @alexluong - do you want this commented out?

alexluong: Yes, it's just a debug log line. It's helpful to have to confirm the timestamps are correct, so I still use it sometimes when developing (both outpost & loadtest), but when running the test, you don't these logs

leggetter: Ah, ok. I think some more of the code that's only used in the logging could also be commented out to avoid confusion.

PR #385, opened 2025-05-15
fix(api): publish response to 202. Tenant response can be 200 or 201

leggetter:
The `202` response from the `/publish` endpoint seems appropriate because it's an async ingestion and no body contents is provided.

@alexbouchardd - let me know what you think?
@alexluong - will this change impact any of the integration tests?

### DIFF
--- a/cmd/e2e/alert_test.go
+++ b/cmd/e2e/alert_test.go
@@ -101,7 +101,7 @@ func (suite *basicSuite) TestConsecutiveFailuresAlert() {
 			}),
 			Expected: APITestExpectation{
 				Match: &httpclient.Response{
-					StatusCode: http.StatusOK,
+					StatusCode: http.StatusAccepted,
 				},
 			},
 		})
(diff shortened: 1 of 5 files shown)


alexbouchardd (approved):

alexluong:
> will this change impact any of the integration tests?

I'll run the test suite tomorrow and push any fixes if necessary!

alexluong:
> The 202 response from the /publish endpoint seems appropriate because it's an async ingestion and no body contents is provided.

Btw, one thing I've been thinking about is whether the `/publish` endpoint should return the event ID? If the user doesn't specify the publish ID, it could be helpful?

The only concern is that they won't be able to query it until the event is attempted at least once. Until then, the event is not stored in the DB yet, and therefore the event retrieve endpoint will return 404.

alexluong:
Alright I'll push a commit for the return payload

```
{
  "id": "..."
}
```

PR #406, opened 2025-05-26
feat: auto generate configuration docs from code to keep docs in sync

leggetter:
See contributing/.plans/2025-05-26-config-documentation-synchronization.md

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -41,21 +41,21 @@ type Config struct {
 	configPath string // stores the path of the config file used
 
 	Service       string              `yaml:"service" env:"SERVICE" desc:"Specifies the service type to run. Valid values: 'api', 'log', 'delivery', or empty/all for singular mode (runs all services)." required:"N"`
-	LogLevel      string              `yaml:"log_level" env:"LOG_LEVEL" desc:"Defines the verbosity of application logs. Common values: 'trace', 'debug', 'info', 'warn', 'error'." required:"Y"`
-	AuditLog      bool                `yaml:"audit_log" env:"AUDIT_LOG" desc:"Enables or disables audit logging for significant events." required:"Y"`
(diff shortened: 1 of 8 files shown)


internal/config/config.go
```diff
@@ -152,7 +153,11 @@ func (c *Config) InitDefaults() {
 	c.Destinations = DestinationsConfig{
 		MetadataPath: "config/outpost/destinations",
 		Webhook: DestinationWebhookConfig{
-			HeaderPrefix: "x-outpost-",
+			HeaderPrefix:             "x-outpost-",
+			SignatureContentTemplate: "{{.Timestamp.Unix}}.{{.Body}}",
+			SignatureHeaderTemplate:  "t={{.Timestamp.Unix}},v0={{.Signatures | join \",\"}}",
+			SignatureEncoding:        "hex",
+			SignatureAlgorithm:       "hmac-sha256",
```
leggetter, line 160: Note that this is a logic change that has crept in. Need to check whether we should remove this. cc: @alexluong

alexluong (approved):
Hey @leggetter, thanks for taking the initiative to set this up. The codegen command was interesting and seems to work quite well. Overall, the PR looks great and doesn't seem to affect the functionality. All test passing right now. We can merge this as is. I do have 2 notes:

1. The docgen command logic is quite complicated (AST, ahhh). I didn't do a full audit but I did run it locally and seems like it works well. My only feedback is I think we could consider a different approach that is a bit simpler.

My understanding is we're parsing files within the `internal/config` package and using AST to detect config structs and parse through it.

I wonder if it's easier to use the `reflect` package to traverse the main `Config` struct. I'll share a snippet below, maybe it can be helpful if you want to explore this approach. Happy to just merge this for now and reiterate later.

2. One semi-unrelated issue is the webhook header. I'll take a look later but if you have time, I'd appreciate it if you can help verify too. I think I changed the behavior of the signature where the default header "X-Outpost-Timestamp" is in epoch timestamp in millisecond, but in the "X-Outpost-Signature", the timestamp there is in second. Because of that, if you implement verification, it won't match correctly unless you convert ms into second first.

PR #409, opened 2025-05-29
chore: Include event data & metadata in alert payload

### LINKED ISSUE #403: Feature Request: Include original webhook topic and payload in `alert.consecutive-failure` events of alerts system**
Yes. We're using the "Alerts" mechanism to monitor webhook delivery failures. However, the payload received in `alert.consecutive-failure` events does not include the exact topic that triggered the failed webhook request (e.g., `user.created`, `user.deleted`, etc.). Additionally, the payload does not include the original request body that failed to be delivered — it only includes the response from the failed attempt. This makes it difficult to debug or understand the context of the failure.

We would like the alert payload to include:
1. A new field that clearly indicates the topic of the webhook event that triggered the delivery failure.
2. The original payload (`body`) that was attempted to be sent to the destination, ideally under a field such as `"original_payload"`.

Currently, we're attempting to correlate failures manually based on the destination URL and timing, but this is unreliable and doesn't scale well. Including the topic and original request payload would make troubleshooting significantly more effective and deterministic.

**Additional context**  
Example of the desired structure in the alert payload:

```json
{
  "topic": "alert.consecutive-failure",
  "timestamp": "2025-05-23T13:26:17.327161906Z",
  "data": {
    "max_consecutive_failures": 5,
    "consecutive_failures": 3,
    "will_disable": false,
    "destination": {
      "id": "5a9fbbe5-432b-44f4-bde9-b2f55a5d44c1",
      "tenant_id": "10089",
      "type": "webhook",
      "topics": ["*"],
      ...
    },
    "webhook_topic": "user.created",
    "original_payload": {
      "user_id": "1234"
    },
    "response": {
      "status": 500,
      "data": {
        "some": "value"
      }
    }
  }
}

### PR DESCRIPTION
**Pull request by alexluong**
resolves #403 

This PR includes an `event` object within the alert payload with some content from the event triggering the alert.

```json
{
  "topic": "alert.consecutive-failure",
  "timestamp": "2025-05-29T05:07:09.269672003Z",
  "data": {
    "event": {
      "topic": "user.created",
      "metadata": {} // ...
      "data": {} // ...
    },
    "max_consecutive_failures": 3,
    "consecutive_failures": 3,
    "will_disable": false,
    "destination": {}, // ...
    "data": {
      "body": "{\"success\":false,\"verified\":false,\"payload\":{\"user_id\":\"userid\"}}",
      "status": 400
    }
  }
}
```

Since this is an addition to the payload, I'd consider this a non-breaking change. With that said, we can also add an opt-in configuration for this (`alert.include_event`) to maintain existing behavior. I do think this is probably safe so maybe we don't need that step, but do let me know if you think it would be helpful!

### DIFF
--- a/internal/alert/notifier.go
+++ b/internal/alert/notifier.go
@@ -42,13 +42,20 @@ func NotifierWithBearerToken(token string) NotifierOption {
 	}
 }
 
+type AlertedEvent struct {
+	Topic    string                 `json:"topic"`    // event topic
+	Metadata map[string]string      `json:"metadata"` // event metadata
+	Data     map[string]interface{} `json:"data"`     // event payload
+}
+
 // ConsecutiveFailureData represents the data needed for a consecutive failure alert
 type ConsecutiveFailureData struct {
+	Event                  AlertedEvent           `json:"event"`
 	MaxConsecutiveFailures int                    `json:"max_consecutive_failures"`
 	ConsecutiveFailures    int                    `json:"consecutive_failures"`
 	WillDisable            bool                   `json:"will_disable"`
 	Destination            *models.Destination    `json:"destination"`
-	Data                   map[string]interface{} `json:"data"`
+	Data                   map[string]interface{} `json:"data"` // alert data, i.e. error message, etc.
 }
 
 // ConsecutiveFailureAlert represents an alert for consecutive failures
--- a/internal/alert/monitor.go
+++ b/internal/alert/monitor.go
@@ -147,6 +147,11 @@ func (m *alertMonitor) HandleAttempt(ctx context.Context, attempt DeliveryAttemp
 	}
 
 	alert := NewConsecutiveFailureAlert(ConsecutiveFailureData{
+		Event: AlertedEvent{
+			Topic:    attempt.DeliveryEvent.Event.Topic,
+			Metadata: attempt.DeliveryEvent.Event.Metadata,
+			Data:     attempt.DeliveryEvent.Event.Data,
+		},
 		MaxConsecutiveFailures: m.autoDisableFailureCount,
 		ConsecutiveFailures:    count,
 		WillDisable:            m.disabler != nil && level == 100,


internal/alert/notifier.go
```diff
@@ -42,13 +42,20 @@ func NotifierWithBearerToken(token string) NotifierOption {
 	}
 }
 
+type AlertedEvent struct {
+	Topic    string                 `json:"topic"`    // event topic
```
leggetter, line 47: should this have the event ID as it is a resource within Outpost?

leggetter:
> @leggetter yes I agree with this but wasn't sure what's the best idea here. I think the `data.data` specifically is probably the easiest thing to clarify. I did add comments in the code since I was confused myself haha.
> 
> Any suggestion from your side?

@alexluong is `data.data` the response from the endpoint that's failing to ingest the event? So, how about `destination_error`? A bit verbose, but clear and no more verbose than some of the other properties. Was going to suggest `_response` but `_error` seems to fit all destination types better.

Suggestion we use an `_` underscore in the topic as it makes it a bit easier to reference (using `.` notation).

```diff
{
- "topic": "alert.consecutive-failure",
+ "topic": "alert.consecutive_failure",
  "timestamp": "2025-05-29T05:07:09.269672003Z",
  "data": {
    "event": {
+     "id": "evt_id"
      "topic": "user.created",
      "metadata": {} // ...
      "data": {} // ...
    },
    "max_consecutive_failures": 3,
    "consecutive_failures": 3,
    "will_disable": false,
    "destination": {}, // ...
-    "data": {
+   "destination_error": {
      "body": "{\"success\":false,\"verified\":false,\"payload\":{\"user_id\":\"userid\"}}",
      "status": 400
    }
  }
}
```

alexbouchardd:
It's really the `destination_response` and not `destination_error` that the object is representing?

We call it "deliveries" in the rest of the product and so `delivery_response` also seems like relevant terminology.

> Since this is an addition to the payload, I'd consider this a non-breaking change. With that said, we can also add an opt-in configuration for this (alert.include_event) to maintain existing behavior. I do think this is probably safe so maybe we don't need that step, but do let me know if you think it would be helpful!

We are in beta now is the time for breaking changes

leggetter:
> It's really the destination_response and not destination_error that the object is representing?

One for @alexluong to answer. Different destinations behave in different ways. Is it always strictly a response or sometimes an error for the SDK we're using?

I'm good with the `delivery_` prefix.

alexluong:
@leggetter updated with everything you suggested above

- [x] include `event.id` in the payload
- [x] change topic name to `alert.consecutive_failure`
- [x] change data to `delivery_response`. I think `delivery_error` works too but I think response is a better name in case in the future we want to trigger alerts for non-error cases. I think response & error are sort of used interchangeably, especially in the "delivery" context. It's also "response" in the dashboard as well.

PR #410, opened 2025-05-29
chore: destawssqs target_url

alexluong:
This PR supports destawssqs to use the target_url field which enable the dashboard to use the `a` tag for the queue target.

CleanShot 2025-05-29 at 12 27 40@2x

One difference between the Hookdeck destination which also uses the target_url mechanism is that the Hookdeck target_url is a link to the Hookdeck dashboard, whereas AWS SQS target_url is just the queue URL which is not something you want to visit from the browser. Because of that, I'm not 100% sure if this is necessary. Since it's a small change, I figured I'd create a PR and we can evaluate / discuss this here.

cc @alexbouchardd @leggetter

### DIFF
--- a/internal/destregistry/providers/destawssqs/destawssqs.go
+++ b/internal/destregistry/providers/destawssqs/destawssqs.go
@@ -208,6 +208,6 @@ func ParseQueueURL(queueURL string) (baseURL string, region string, err error) {
 func (d *AWSSQSDestination) ComputeTarget(destination *models.Destination) destregistry.DestinationTarget {
 	return destregistry.DestinationTarget{
 		Target:    destination.Config["queue_url"],
-		TargetURL: "",
+		TargetURL: destination.Config["queue_url"],
 	}
 }


alexbouchardd:
The target_url in this context would have to be the URL in the AWS console. If we can build the console URL from the queue URL it could be relevant. I wouldn't support the URL of the SQS resource itself since it effectively doesn't link anywhere useful

alexluong:
@alexbouchardd that makes a lot more sense. We can probably construct the console URL I think, not 100% sure.

One thing to note is that we technically support a SQS-compat service, in which case the console URL won't work correctly. This is an edge cade and maybe not a usage we want to consider or support. Any thoughts here?

alexbouchardd:
> This is an edge cade and maybe not a usage we want to consider or support. Any thoughts here?

You should be able to tell that from the URL structure but given the extra complexity here it's low prio IMO

alexluong:
Yeah it's not too complex, I'll push a commit for this soon. On the dashboard, should the copy button copy the name or the URL itself? Also it could be a bit unintuitive that the link content is a URL (the queue URL), but when clicking on it, it opens a different link (the console) instead.

@alexbouchardd  cc @leggetter

alexbouchardd:
The copy should copy the value that is shown, not the URL. I don't think it's a problem that clicking it opens something else as although it's a queue URL it's really just a resource reference in the context of AWS

alexluong:
Great, in that case this PR is ready to go then!

PR #425, opened 2025-06-09
feat: Support Azure ServiceBus PubSub for internalmq

alexluong:


internal/config/mqconfig_azure.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (c *AzureServiceBusConfig) ToQueueConfig(ctx context.Context, queueType string) (*mqs.QueueConfig, error) {
+	if !c.IsConfigured() {
+		return nil, nil
+	}
+
+	// connectionString, err := c.getConnectionString(ctx)
+	// if err != nil {
+	// 	return nil, err
+	// }
```
alexbouchardd: ?!

alexluong: Right, thanks for noticing this, let me explain. From my understanding, there are 2 authentication mechanisms for Azure Service:

- connection string
- identity-based

In my first implementation, I couldn't quite get the identity-based approach to work so I implemented this version with the connection string. However, I figured out the right role so this mechanism is no longer necessary.

This is sort of around roles & permissions. If there's the right role, "Azure Service Bus Data Owner", then we don't need the connection string. However, if the provided service principal only has roles on the management plane instead of data plane, then we do need the connection string mechanism.

Ultimately, if we communicate to the user what permission we need, then we don't need this mechanism anymore. In that case, we can probably just remove this code.

leggetter: @alexluong - how about leaving a comment here that outlines this for anybody coming to this in the future?

leggetter:
@alexluong - following the setup instructions and then running Outpost locally with `make up`, I get the following in the `api` service logs:

```
running...
{"level":"info","ts":1749544317.5334902,"caller":"app/app.go:47","msg":"starting outpost","config_path":".env","service":"api"}
{"level":"info","ts":1749544317.7974298,"caller":"app/app.go:197","msg":"no migrations applied","version":3}
failed to declare topic: failed to create topic outpost-delivery: PUT https://management.azure.com/subscriptions/21f864a1-c750-44d6-b40a-bf62a33541aa/resourceGroups/outpost-rg/providers/Microsoft.ServiceBus/namespaces/outpost-servicebus-271b4344/topics/outpost-delivery
--------------------------------------------------------------------------------
RESPONSE 429: 429 Too Many Requests
ERROR CODE: MessagingGatewayTooManyRequests
--------------------------------------------------------------------------------
{
  "error": {
    "code": "MessagingGatewayTooManyRequests",
    "message": "SubCode=40900. Conflict. You're requesting an operation that isn't allowed in the resource's current state. To know more visit https://aka.ms/sbResourceMgrExceptions.  ExceptionDescription: sb://outpost-servicebus-271b4344.servicebus.windows.net/outpost-delivery. TrackingId:49463282-fa78-475b-8110-355879cf2d3b_G13, SystemTracker:outpost-servicebus-271b4344.servicebus.windows.net:outpost-delivery, Timestamp:2025-06-10T08:32:01"
  }
}
--------------------------------------------------------------------------------

Process Exit with Code: 1
```

Something to do with permissions? Or a timing issue?

```
failed to declare topic: failed to create topic outpost-delivery: PUT https://management.azure.com/subscriptions/21f864a1-c750-44d6-b40a-bf62a33541aa/resourceGroups/outpost-rg/providers/Microsoft.ServiceBus/namespaces/outpost-servicebus-271b4344/topics/outpost-delivery
```

alexluong:
@leggetter yes I noticed some issues on the first run, most likely because the 3 services are trying to create the same pubsub at the same time. I don't have a solution yet. Locally what I do is I run the API node first, then after the infra is provisioned, I run the other.

Not sure why we haven't run into this with the other mqs. We may need to implement a mechanism in the infra logic to only allow 1 provision to happen at once. Created #427 here so we can track this issue.

internal/config/mqconfig_azure.go
alexluong: I cleaned up this file a bit and removed this snippet, commented out the unused methods / values, and included a comment to explain the different Azure auth mechanism. Hope that's sufficient to explain the thought process here.

leggetter:
@alexluong - configured locally and tested.

I did come across an error that I think is unrelated to this PR when running `examples/demos/nodejs` / `npm run publish-api`:

```
fatal error: all goroutines are asleep - deadlock!
⠙
goroutine 1 [chan receive]:
github.com/evanw/esbuild/internal/helpers.(*ThreadSafeWaitGroup).Wait(...)
        github.com/evanw/esbuild/internal/helpers/waitgroup.go:36
main.runService.func2()
        github.com/evanw/esbuild/cmd/esbuild/service.go:114 +0x8c
main.runService(0x1)
        github.com/evanw/esbuild/cmd/esbuild/service.go:160 +0x4b4
main.main()
        github.com/evanw/esbuild/cmd/esbuild/main.go:236 +0x8e4

goroutine 34 [chan receive]:
main.runService.func1()
        github.com/evanw/esbuild/cmd/esbuild/service.go:98 +0x40
created by main.runService
        github.com/evanw/esbuild/cmd/esbuild/service.go:97 +0x1a0

goroutine 35 [chan receive]:
main.(*serviceType).sendRequest(0x1400010de60, {0x104b5b4e0, 0x140003a04e0})
        github.com/evanw/esbuild/cmd/esbuild/service.go:192 +0x11c
main.runService.func3()
        github.com/evanw/esbuild/cmd/esbuild/service.go:125 +0x38
created by main.runService
        github.com/evanw/esbuild/cmd/esbuild/service.go:122 +0x308
```

I've also updated the config reference and created a guide to setting up Azure Service Bus based on the details you provided in this PR.

alexluong:
@leggetter this is an issue with `esbuild` which uses Go under the hood. I don't think it's related to Outpost itself. Can you try investigating esbuild and maybe reinstall deps (npm rm -rf node_modules && npm install) to see if it resolves this issue?

leggetter:
@alexluong - if required, let's capture these in an issue for later:

> * [ ]  set up emulator for integration testing with Azure
> * [ ]  _light_ load test to make sure the Azure integration (pub, sub) can handle moderate load

---

> @leggetter this is an issue with esbuild which uses Go under the hood. I don't think it's related to Outpost itself. Can you try investigating esbuild and maybe reinstall deps (npm rm -rf node_modules && npm install) to see if it resolves this issue?

Aha! Ok, this shouldn't block this from being merged. I'll investigate later.

PR #434, opened 2025-06-25
chore: Setup Azure SB Emulator for testing & local development

alexluong:


internal/config/mqconfig_azure.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 type AzureServiceBusConfig struct {
+	// Using AzureSB with ConnectionString will skip infra management
+	ConnectionString string `yaml:"connection_string" env:"AZURE_SERVICEBUS_CONNECTION_STRING" desc:"Azure Service Bus connection string" required:"N"`
```
alexluong, line 12: I want to call attention to this config a bit. This should be "dev-only" config as it's not recommended in production.

Outpost is responsible for managing its own mq infra. Using Azure connection string will bypass this mechanism. This could work for now if the operator is comfortable managing their own resources, and most importantly follow along Outpost development to make changes if necessary.

Ideally, I'd like to hide this from config away from users if possible. I haven't explored the config docgen much, do you have a suggestion here? @leggetter

internal/util/testinfra/azure.go
```diff
@@ -0,0 +1,82 @@
+// Azure Service Bus Testing Infrastructure
```
alexluong, line 1: Also calling this out as a significant limitation of Azure ServiceBus emulator

build/dev/deps/compose.yml
```diff
@@ -63,7 +63,9 @@ services:
   rabbitmq:
     image: rabbitmq:3-management
     ports:
-      - 5672:5672
+      # Reserve 5672 for AzureSB Emulator which doens't support custom ports.
+      # We may not need to expose 5672 at all if we run the full local dev env with Docker Compose.
+      # - 5673:5672
```
alexluong, line 68: Another quirk. Basically AzureSB emulator requires port 5672 to work, so it's stealing that port from our RabbitMQ deps

internal/config/mqconfig_azure.go
leggetter: @alexluong - we could add another attribute: `...required:"N" hidden: true`?

alexluong: Yes, that should work!

alexbouchardd (approved):

PR #456, opened 2025-08-04
Implement S3 Support

eberle1080:
This PR implements S3 support. It stores messages as objects in S3 and allows certain attributes to be configured, such as:
* Bucket name
* Region
* Object prefix + suffix
* Some control over object dynamic naming (admittedly this could be more flexible)
* Storage class


internal/destregistry/providers/destawss3/destawss3.go
```diff
@@ -0,0 +1,331 @@
...
+)
+
+// AWSS3Config is the configuration for an S3 destination
+type AWSS3Config struct {
+	Bucket           string
+	Region           string
+	KeyPrefix        string
+	KeySuffix        string
+	IncludeTimestamp bool
+	IncludeEventID   bool
```
alexluong: Should we consider using a template language for this?

We use JMESPath for [AWS Kinensis partition key](https://github.com/hookdeck/outpost/blob/3ba0d610222549bb66d347c03dee722a98b3a86d/internal/destregistry/providers/destawskinesis/destawskinesis.go#L186-L219). I believe JMESPath is decently adopted within the AWS ecosystem.

internal/destregistry/providers/destawss3/destawss3.go
```diff
@@ -0,0 +1,331 @@
...
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	awssdk "github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	awscreds "github.com/aws/aws-sdk-go-v2/credentials"
```
alexluong, line 15: small nit but is there a reason why we alias these import statements? Do let me know if it's convention within the Go AWS SDK. I'm curious to learn more!

internal/destregistry/providers/destawss3/destawss3.go
```diff
@@ -0,0 +1,331 @@
...
+	includeEventID := true
+	if eidStr, ok := destination.Config["include_event_id"]; ok {
+		includeEventID = eidStr == "true" || eidStr == "on"
+	}
+
+	sc := destination.Config["storage_class"]
+
+	_, err := parseStorageClass(sc)
+	if err != nil {
+		return nil, nil, fmt.Errorf("invalid storage class %q: %w", sc, err)
```
alexluong: we should return `destregistry.NewErrDestinationValidation` if possible, see `destwebhook` for an example

internal/destregistry/providers/destawss3/destawss3.go
```diff
@@ -0,0 +1,331 @@
...
+		}, nil
+}
+
+// ComputeTarget returns a human-readable target description for the S3 destination
+func (p *AWSS3Provider) ComputeTarget(destination *models.Destination) destregistry.DestinationTarget {
+	bucket := destination.Config["bucket"]
+	region := destination.Config["region"]
+	return destregistry.DestinationTarget{
+		Target:    fmt.Sprintf("%s in %s", bucket, region),
+		TargetURL: "",
```
alexluong: Can we construct the AWS console URL of this S3 bucket?

internal/destregistry/providers/destawss3/destawss3.go
```diff
@@ -0,0 +1,331 @@
...
+	checksumSha256, err := p.getChecksums(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to compute checksum: %w", err)
+	}
+
+	return &s3.PutObjectInput{
+		Bucket:            awssdk.String(p.bucket),
+		Key:               awssdk.String(key),
+		Body:              bytes.NewReader(data),
+		Metadata:          event.Metadata,
```
alexluong: We should use the `MakeMetadata` function from the base publisher which includes some Outpost-specific metadata.

internal/destregistry/providers/destawss3/destawss3_format_test.go
```diff
@@ -0,0 +1,71 @@
+package destawss3_test
```
alexluong, line 1: Besides the format & validation test, can we also include the publish test? Outpost provides a standardized publisher test suite that all destinations should pass. Happy to share more or elaborate more on this if you have any questions.

examples/docker-compose/compose-awss3.yml
```diff
@@ -0,0 +1,41 @@
...
+
+  api:
+    depends_on:
+      aws:
+        condition: service_healthy
+    environment:
+      - AWS_S3_ENDPOINT=http://aws:4566
+      - AWS_S3_REGION=us-east-1
+      - AWS_S3_ACCESS_KEY_ID=test
+      - AWS_S3_SECRET_ACCESS_KEY=test
```
alexluong, line 23: This is an interesting approach. So we're embedding the env variables so that the SDK will always connect to LocalStack instead of real AWS environment?

I think it's a pretty neat approach. So far, when implementing AWS destinations, we support a hidden config value of "endpoint" to configure this. I'm not sure if that is something we should consider supporting here, mainly for consistency sake. But I do like this idea a lot. But I don't imagine this works well during the local dev but only during the examples? Any tips from your side?

internal/destregistry/providers/destawss3/destawss3.go
alexluong: What do you think if we replace these config options with `KeyTemplate`?

In Kinesis, the input of the template is the same as the full `payload` document. In Kinesis, there's a Outpost-specific configuration so the payload could either be `data` or be a map of both `data` and `metadata`.

We can do something similar-ish here for consistency sake?

internal/destregistry/providers/destawss3/destawss3.go
eberle1080: readability mostly. Just slightly nicer to read. But no, not a convention.

internal/destregistry/providers/destawss3/destawss3.go
alexluong: Small note since I'm thinking on this a bit further. There's a small nuance or decision around the input of the template function here. I'll share the Kinesis destination approach first and discuss some pros / cons.

In Kinesis, the event payload is the payload of the template function for its partition key. The reason is the structure is something the end-user should expect, so it's something they can use the customize the partition key accordingly. Furthermore, there's an Outpost-specific config (not destination config) where the operator can decide to include metadata in the Kinesis event payload or not.

Back to S3, we can consider something similar, using `event.data` as the payload. However, it can potentially miss critical metadata that the user may want to use to customize the template such as `event-id`, `timestamp`, or `topic`.

What do you think is the right balance of information / customization while maintaining an API surface or the DX that makes sense?

alexluong: For example, say this is the event

```json
{
    "id": "evt_1",
    "tenant_id": "ten_1",
    "topic": "user.created",
    "eligible_for_retry": false,
    "metadata": {
        "my_metadata": "metadata",
    },
    "data": {
        "my_data": "data"
    }
}
```

In the current S3 implementation, this would be the payload of the stored event:

```json
{
  "my_data": "data"
}
```

I'm concerned this may not be sufficient for the end-users as they may not be able to meaningfully come up with a template that makes sense for them.

We can consider something like this:

```json
{
  "metadata": {
    "my_metadata: "metadata",
    // system metadata
    "event-id": "evt_1",
    "topic": "user.created",
    "timestamp": 123456,
  },
  "data": {
    "my_data": "data"
  }
}
```

This would have enough information for the user, but it's a bit unintuitive and unfamiliar to the end users as they may not understand what is available within `metadata` itself.

alexbouchardd: What I'd explore is to set a default value for the field, which can also be displayed in the UI with a sensible default. That educates the user about the shape and properties of the field. What's the shape of the data actually saved to S3? I'd expect it to include the metadata? Ideally the data shape used for the template should match the data persisted

alexluong: In the PR, the shape seems to be just event.data

alexbouchardd: I'd argue it should include the metadata 

```
{
  "metadata": {
    "my_metadata: "metadata",
    "event-id": "evt_1",
    "topic": "user.created",
    "timestamp": 123456,
  },
  "data": {
    "my_data": "data"
  }
}
```

eberle1080: I think that all makes sense. I'm good with that. Do you think it might make sense to choose between "data + metadata" vs "raw data only"? For us I think either approach would work but I wonder if some might be more opinionated about the shape of their data.

internal/destregistry/providers/destawss3/destawss3_format_test.go
eberle1080: yeah if you could link to an example of that it would be helpful. Happy to add this.

internal/destregistry/providers/destawss3/destawss3.go
alexbouchardd: Given the way most vendor structure their webhooks, the metadata (headers) often contain fairly essential pieces of information, such as the event topic. I would lean toward always including it?

leggetter: +1 to including the metadata

alexluong: I proceeded with the implementation for the key template. Here's the input for the template:

```json
  {
    "data": {
      // Event data - preserves original field names/casing
      "userId": "123",
      "orderAmount": 99.99
    },
    "metadata": {
      // System metadata + user metadata - preserves original casing
      "event-id": "evt_123",
      "topic": "order.created",
      "timestamp": "2024-01-15T10:30:00Z",
      "customField": "value"
    },
    "time": {
      // Pre-parsed UTC time components (all strings for consistency)
      "year": "2024",
      "month": "01",
      "day": "15",
      "hour": "10",
      "date": "2024-01-15",
      "datetime": "2024-01-15T10:30:00",
      "unix": "1705314600",
      "rfc3339_nano": "2024-01-15T10:30:00.000000000Z"
    }
  }
```

The reasons I'm including the pre-parsed `time` are
- JMESPath doesn't provide built-in time parsing functions and it's fairly
- S3 key naming pattern often involves date partitioning

```
s3://bucket/year=2024/month=01/day=15/event-123.json
s3://bucket/2024/01/15/logs/event-123.json
s3://bucket/data/2024-01-15/event-123.json
```

internal/destregistry/providers/destawss3/destawss3_format_test.go
```diff
@@ -0,0 +1,433 @@
...
+// TestAWSS3Publisher_Format_LegacyPatterns demonstrates how to replicate the old
+// configuration options (key_prefix, key_suffix, include_timestamp, include_event_id)
+// using the new JMESPath template system.
+//
+// The new template system is much more flexible and allows:
+// - Date-based partitioning: `join('/', ['year=', time.year, 'month=', time.month, 'day=', time.day, metadata."event-id"])`
+// - Topic-based organization: `join('/', [metadata.topic, time.date, metadata."event-id"])`
+// - Data-driven paths: `join('/', ['users', data.user_id, 'events', time.unix])`
+// - Complex nested structures: `join('/', [metadata.env, metadata.region, time.year, metadata.topic, data.order_id])`
+func TestAWSS3Publisher_Format_LegacyPatterns(t *testing.T) {
```
alexluong, line 335: I provided a test suite of some example patterns in the past and how it's implemented with JMESPath template.

Example templates

```
join('', ['events/', metadata."event-id", '.json'])
# events/event-123.json

join('', [time.rfc3339_nano, '_', metadata."event-id", '.json'])
# 2024-01-02T03:04:05.000000000Z_event-123.json

metadata."event-id"
# event-123

time.rfc3339_nano
# 2024-01-02T03:04:05.000000000Z

join('', ['logs/', time.year, '/', time.month, '/', metadata.topic, '/', time.day, '_', metadata."event-id", '.gz'])
# logs/2024/01/user.created/02_event-123.gz
```

I must admit, the template looks a bit more complex than I'd like. But given JMESPath should be decently adopted within the AWS community, I think this is a good starting point.

An idea to improve this in the future is to provide with Outpost operator the ability to configure which template engine they want to provide (or the shape of the input).

PR #481, opened 2025-09-06
chore(api): log /api/v1/healthz requests at debug level instead of info to reduce noise

obazoud:


internal/services/api/logger_middleware.go
```diff
@@ -60,7 +60,11 @@ func LoggerMiddlewareWithSanitizer(logger *logging.Logger, sanitizer *RequestBod
 				hub.CaptureException(getErrorWithStackTrace(c.Errors.Last().Err))
 			}
 		} else {
-			logger.Info("request completed", fields...)
+			if c.Request.URL.Path == "/api/v1/healthz" {
```
alexluong: Small nit, would it be better to check "string ending with /healthz and starting with /api" so it works for new versions in the future.

Not a big deal tho, just let me know, we can merge as-is.

alexluong (approved):

alexluong (approved):
looks great, thanks for your contribution!

PR #482, opened 2025-09-06
chore(build): Copy go mod files first to leverage Docker layer cachin…

obazoud:
…g for dependencies

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,9 +1,9 @@
 # Stage 0
 # Build the binary
 FROM golang:1.23-alpine
+WORKDIR /app
 COPY go.mod go.sum ./
 RUN go mod download
-WORKDIR /app
 COPY . .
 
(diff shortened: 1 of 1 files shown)


alexbouchardd (approved):

alexluong (commented):
Similar question to [the message in another thread](https://github.com/hookdeck/outpost/pull/483#issuecomment-3263447134). Since I personally don't use this Dockerfile, I just want to understand the motivation.

Potentially, I think it may make sense to simply remove this Dockerfil instead if you can continue your workflow without. With that in mind, happy to proceed and support this if it helps.

I just think this Dockerfile is nowhere close to production-ready, and I want to avoid spending time maintaining an unused Dockerfile if we can avoid it.

obazoud:
I’ve explained in the [other thread](https://github.com/hookdeck/outpost/pull/483#issuecomment-3263601030) how I use this Dockerfile in my local workflow to test my changes.

Issue #490, opened 2025-09-16
SDK and OpenAPI mismatch with raw HTTP response fields


obazoud:
**Describe the bug**
There is a discrepancy between the attributes returned in the SDK (and OpenAPI spec) versus the raw HTTP response from the events endpoints (tenant and destination):

When requesting:
GET `https://myoutpost.com/api/v1/<tenant id>/destinations/<destination id>/events`

The response body is like:
```
{
  count: x,
  data: [
    {
      id: '<event id>',
      tenant_id: '<tenant id>',
      destination_id: '<destination id>',
      topic: '<my topic>',
      eligible_for_retry: true,
      time: '2025-09-16T15:27:57.720776Z',
      metadata: {},
      data: {...},
      status: 'success'
    },
    ...
  ],
  next: '',
  prev: ''
}
```

But in `openapi.yaml`, the array should consist of Event components (`#/components/schemas/Event`), which are defined to include:
- id
- destination_id
- topic
- time
- successful_at
- metadata
- data

However, the raw HTTP response includes fields not present in the SDK/OpenAPI, and vice versa:
- Missing fields in the SDK/OpenAPI: `tenant_id`, `status`, `eligible_for_retry`
- Missing fields in HTTP: `successful_at`

The portal uses raw HTTP, which is why it works, but SDKs and OpenAPI-based integrations miss some fields.

**To Reproduce**
1. Make the above GET request to `/events` endpoints.
2. Compare the response fields to the OpenAPI Event schema and SDK models.

**Expected behavior**
The SDK and OpenAPI schema should match the raw HTTP response.

**Screenshots**
N/A

**Additional context**
N/A

obazoud:
GET `/<tenant_id>/events/<event_id>/deliveries` can take a `destination_id` parameter to [filter the delivery attempts](https://github.com/hookdeck/outpost/blob/main/internal/services/api/log_handlers.go#L161-L164).

However, this [parameter is missing](https://github.com/hookdeck/outpost/blob/main/docs/apis/openapi.yaml#L2067-L2109) from the OpenAPI specification.

obazoud:
GET `/<tenant_id>/events/<event_id>/deliveries` : discrepancy between [OpenAPI](https://github.com/hookdeck/outpost/blob/main/docs/apis/openapi.yaml#L1008-L1029) and the [raw HTTP response](https://github.com/hookdeck/outpost/blob/main/internal/services/api/log_handlers.go#L175-L184)

PR #491, opened 2025-09-17
fix(sdk): add missing status #490

obazoud:


leggetter:
Fix included in https://github.com/hookdeck/outpost/pull/518 along with a way to help us test SDKs against a running version of Outpost and identify any discrepancies.

Issue #306, opened 2025-04-02
Improve Health Check and Service Startup Flow


alexluong:
Currently, services can appear "ready" while having connection issues to dependencies (e.g. Redis timeout). We should enhance the health check and startup flow to properly verify all critical dependencies are connected before marking a service as ready. This should include proper readiness probes for k8s, connection retries with backoff, and clear error reporting when dependencies aren't available. This will prevent silent failures where services appear healthy but aren't fully operational.

Specifically, for API service, we run a few different things: API router, publishmq consumer, retrymq consumer. If one of the consumer fails, there's an error log line but we don't escalate it further.

PR #558, opened 2025-11-10
feat: healthcheck & worker refactor

### LINKED ISSUE #555: Consumer Exits Permanently on Errors without Auto-Recovery
When the SQS consumer encounters an error, the consumer loop exits permanently and never restarts. The service appears healthy (container running, HTTP responding) but stops processing all messages from the queue. This causes silent failures where new messages accumulate but are never processed until someone manually restarts the service.

We experienced this when a message with malformed data caused an unmarshal error (`json: cannot unmarshal string into Go struct field PublishedEvent.data of type map[string]interface{}`). The handler attempted to Nack the message, which failed due to missing `sqs:ChangeMessageVisibility` IAM permission. This caused the consumer to exit **silently** and the service appeared healthy, but stopped processing messages until manual restart.

**To Reproduce**

1. Deploy Outpost with AWS SQS but missing `sqs:ChangeMessageVisibility` permission
2. Message arrives that causes any handler error (malformed data, processing failure, etc.)
3. Handler attempts to Nack the failed message
4. GoCloud driver tries to call `ChangeMessageVisibilityBatch`
5. AWS returns 403 permission denied
6. Error bubbles up from consumer.Run() to startPublishMQConsumer (api/api.go:253)
7. Consumer goroutine logs error and exits permanently
8. Service continues running, appears healthy
9. **All subsequent messages accumulate unprocessed with no alerts or visible indication**

alexluong:
resolves #555, #306

### Health Check Endpoint

This PR updates the `/healthz` endpoint to report the real-time health status of all workers in the application. The endpoint returns:

**Healthy response (200):**
```json
{
  "status": "healthy",
  "timestamp": "2025-01-15T10:30:00Z",
  "workers": {
    "http-server": {"status": "healthy"},
    "retrymq-consumer": {"status": "healthy"},
    "deliverymq-consumer": {"status": "healthy"}
  }
}
```

**Unhealthy response (503):**
```json
{
  "status": "failed",
  "timestamp": "2025-01-15T10:30:00Z",
  "workers": {
    "http-server": {"status": "healthy"},
    "retrymq-consumer": {"status": "healthy"},
    "deliverymq-consumer": {"status": "failed"}
  }
}
```

**Usage:** Configure Kubernetes liveness/readiness probes or load balancer health checks to monitor this endpoint. When any worker fails, the endpoint returns 503, allowing orchestrators to automatically restart the pod/container.

**Note:** For `service=delivery` and `service=log` modes, an HTTP server is started solely to serve the `/healthz` endpoint on the configured API port, allowing health checks even for worker-only services.

### Worker Supervision Architecture

The application now uses a **WorkerSupervisor** pattern to manage long-running workers (HTTP server, MQ consumers, retry scheduler). Key design decisions:

- **Non-cascading failures:** When a worker fails, it's marked as failed but other workers continue running. This keeps the HTTP server and health endpoint available for debugging while the orchestrator detects the failure and restarts the container.
- **Graceful shutdown:** On SIGTERM/SIGINT, the supervisor cancels the context and waits for all workers to finish gracefully, then runs cleanup (database connections, HTTP server shutdown, etc.) before exiting.

This architecture provides visibility into worker health while maintaining availability during partial failures.

### DIFF
--- a/internal/services/builder.go
+++ b/internal/services/builder.go
@@ -128,12 +128,18 @@ func (b *ServiceBuilder) createHTTPServer(router http.Handler) error {
 }
 
 // Cleanup runs all registered cleanup functions for all services.
+// Cleanup is performed in LIFO (last-in-first-out) order to ensure that
+// resources created later (which may depend on earlier resources) are
+// cleaned up before their dependencies.
 func (b *ServiceBuilder) Cleanup(ctx context.Context) {
 	logger := b.logger.Ctx(ctx)
-	for _, svc := range b.services {
+	// Clean up services in reverse order (LIFO)
+	for i := len(b.services) - 1; i >= 0; i-- {
+		svc := b.services[i]
 		logger.Debug("cleaning up service", zap.String("service", svc.name))
-		for _, cleanupFunc := range svc.cleanupFuncs {
-			cleanupFunc(ctx, &logger)
+		// Clean up functions within each service in reverse order
+		for j := len(svc.cleanupFuncs) - 1; j >= 0; j-- {
+			svc.cleanupFuncs[j](ctx, &logger)
 		}
 	}
 }


leggetter:
@alexluong - just want to flag that the payload change is a breaking change and there may be some code that checks for "OK", including in the Outpost repo.

alexluong:
> including in the Outpost repo

Can you reference this? I'm not aware of any usage of this.

Good call regarding the breaking change. Seems we need to issue v0.9.0 for this in that case?

fmvilas:
> Seems we need to issue v0.9.0 for this in that case?

I'd definitely go for 0.9.0.

leggetter:
> > including in the Outpost repo
> 
> Can you reference this? I'm not aware of any usage of this.

@alexluong - I had a quick look, but I can't find anything programmatically checking for `OK`. Another set of eyes would be useful.

I do see the main README mentions the response contents, so that will need to be changed. Worth looking at other docs, too.

alexluong:
@leggetter I looked at `/docs` directory already and don't think we mention the content of the healthz endpoint

updated README

fmvilas:
I also can't find anything else. In the website preview link the API Reference is not updated but I'm assuming this will be automatically updated once we release?

fmvilas (approved):
LGTM 👍

alexluong:
@claude can you review pleaes?

PR #560, opened 2025-11-12
feat: make mqs infra provisioning configurable

### LINKED ISSUE #547: A bug with SQS FIFO and DLQ
The dlq check is naive, and appends `-dlq` to the end of the queue name.

However, fifo queues require to have the `.fifo` suffix, and thereby the final queue name becomes `foo-bar.fifo-dlq` which is an invalid queue name.

https://github.com/hookdeck/outpost/blob/732545e40b5b307aa0c0a85ee26264c415f27909/internal/mqinfra/awssqs.go#L87

I think the correct solution is to allow the end user to configure the name of all queues via config and/or env, and thereby don't need to have this check at all.

alexluong:
resolves #547

introduces `MQS_SHOULD_MANAGE` environment variable (open to changing to a better naming convention) which allows users to self-manage the mqs infra themselves

### DIFF
--- a/internal/infra/infra.go
+++ b/internal/infra/infra.go
@@ -19,7 +19,7 @@ const (
 
 var (
 	// ErrInfraNotFound is returned when infrastructure does not exist and auto provisioning is disabled
-	ErrInfraNotFound = errors.New("infrastructure does not exist and auto provisioning is disabled (MQS_AUTO_PROVISION=false). Please create the required message queues manually or set MQS_AUTO_PROVISION=true to enable auto provisioning")
+	ErrInfraNotFound = errors.New("required message queues do not exist. Either create them manually or set MQS_AUTO_PROVISION=true to enable auto-provisioning")
 )
 
 type Infra struct {


docs/pages/references/configuration.mdx
```diff
@@ -80,6 +80,7 @@ Global configurations are provided through env variables or a YAML file. ConfigM
 | `LOG_MAX_CONCURRENCY` | Maximum number of log writing operations to process concurrently. | `1` | No |
 | `MAX_DESTINATIONS_PER_TENANT` | Maximum number of destinations allowed per tenant/organization. | `20` | No |
 | `MAX_RETRY_LIMIT` | Maximum number of retry attempts for a single event delivery before giving up. Ignored if retry_schedule is provided. | `10` | No |
+| `MQS_SHOULD_MANAGE` | Whether Outpost should create and manage message queue infrastructure. Set to false if you manage infrastructure externally (e.g., via Terraform). Defaults to true for backward compatibility. | `nil` | No |
```
fmvilas: What about this?


```suggestion
| `MQS_AUTO_CREATE` | Whether Outpost should create and manage message queue infrastructure. Set to false if you manage infrastructure externally (e.g., via Terraform). Defaults to true for backward compatibility. | `nil` | No |
```

Or this?


```suggestion
| `MQS_AUTO_MANAGE` | Whether Outpost should create and manage message queue infrastructure. Set to false if you manage infrastructure externally (e.g., via Terraform). Defaults to true for backward compatibility. | `nil` | No |
```

fmvilas (commented):
Suggesting an alternative name. Still looking at the code though.

docs/pages/references/configuration.mdx
fmvilas: Or even this?


```suggestion
| `MQS_AUTO_PROVISION` | Whether Outpost should create and manage message queue infrastructure. Set to false if you manage infrastructure externally (e.g., via Terraform). Defaults to true for backward compatibility. | `nil` | No |
```

internal/app/app.go
```diff
@@ -32,258 +40,223 @@ func New(cfg *config.Config) *App {
 }
 
 func (a *App) Run(ctx context.Context) error {
-	return run(ctx, a.config)
+	if err := a.PreRun(ctx); err != nil {
+		return err
+	}
+	defer a.PostRun(ctx)
+
+	return a.run(ctx)
```
fmvilas, line 48: Love this new lifecycle separation. Looks much cleaner.

fmvilas (approved):
Love this refactor, @alexluong 👏 

It makes things way more readable now. You can't imagine how much I appreciate this being new with this codebase 😄 🙏

alexluong:
@claude review please, thanks

PR #562, opened 2025-11-12
feat: make retrymq monitor loop timeout configurable

### LINKED ISSUE #486: Monitor loop is too aggressive
**Describe the bug**
The monitor loop currently polls Redis at a high frequency (every 100ms), especially in development environments. When using a serverless Redis provider such as Upstash, the quota is quickly reached due to constant polling. If using ElasticCache Serverless, this aggressive polling can also lead to additional costs (few $/month)

From what I understand, this monitor loop is only used for retrying failed deliveries. It would be better to trigger the retry mechanism only when there are actually retries to process, instead of polling continuously.

**To Reproduce**
1. Start the API service
2. Check your Redis logs or monitoring
3. You should see the commands from the receiveMessage LUA Script every 100ms

**Expected behavior**
- Quick fix: the polling interval should be configurable. So in development and with serverless/pay-as-you-go Redis providers, the system should not quickly exhaust the quota or increase cost
- The monitor loop should avoid aggressive polling when there are no retry messages to process
- The retry mechanism should only be triggered when there are retries to process

**Screenshots**
N/A

**Additional context**
N/A

alexluong:
resolves #486

env: `RetryPollBackoffMs`


fmvilas:
Gonna have a look. Fun fact: I was challenging myself trying to solve this same issue today 😄 https://github.com/hookdeck/outpost/tree/feat/configurable-redis-polling.

Issue #480, opened 2025-09-05
Improve Config & secrets handling


alexluong:
**Is your feature request related to a problem? Please describe.**
Inspired by #470, Outpost should be more careful when handling sensitive values. Currently, it's quite easy to accidentally log a secret. Outpost should make it harder to make that type of mistake.

**Describe the solution you'd like**
I think there are 2 ideas and Outpost should consider both.

1. Instead of using normal String type, Outpost should use some sort of "SensitiveString" type or applying [a masker](https://github.com/showa-93/go-mask) of some sort. That way, when developers add logging to a config struct with senstive values, it will not logged by default.
2. Consider Config schema and evaluate whether there's a way to be more intentional about the distinction between application config which drives the behavior of Outpost vs necessary credentials.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #564, opened 2025-11-12
feat: improve startup logging

### LINKED ISSUE #485: Improve startup logging
The application startup should provide more detailed logs to help with configuration and troubleshooting. Currently, when a Redis connection fails, the logs only show EOF or dial tcp://xx.xx.xx.xx, which is not clear enough.

Add explicit and structured logs during startup to make configuration and connection issues easier to diagnose.
It could be:
- Display the path to the configuration file (if present)
- Show the values of environment variables related to Outpost (excluding secrets)
- Log the result of configuration merging and resolution
- Add an explicite logs indicating the success or failure of checks (dlq queues, ...), external connections (e.g., Redis, PostgreSQL, open a subscription, ...)
...

N/A

**Additional context**
N/A

fmvilas:
Fixes #485 

Logs a configuration summary at startup with the following info:

```json
{
  "msg": "starting outpost",
  "service": "api",
  "config_file_path": ".env",
  "log_level": "info",
  "audit_log": true,
  "deployment_id": "",
  "topics": [
    "user.created",
    "user.updated",
    "user.deleted"
  ],
  "organization_name": "",
  "http_user_agent": "",
  "api_port": 3333,
  "api_key_configured": true,
  "api_jwt_secret_configured": true,
  "gin_mode": "release",
  "aes_encryption_secret_configured": true,
  "redis_host": "redis",
  "redis_port": 6379,
  "redis_password_configured": true,
  "redis_database": 0,
  "redis_tls_enabled": false,
  "redis_cluster_enabled": false,
  "postgres_configured": true,
  "postgres_host": "postgres:5432",
  "mq_type": "rabbitmq",
  "publish_max_concurrency": 1,
  "delivery_max_concurrency": 1,
  "log_max_concurrency": 1,
  "retry_schedule": [],
  "retry_interval_seconds": 30,
  "retry_max_limit": 10,
  "max_destinations_per_tenant": 20,
  "delivery_timeout_seconds": 5,
  "publish_idempotency_key_ttl": 3600,
  "delivery_idempotency_key_ttl": 3600,
  "log_batch_threshold_seconds": 10,
  "log_batch_size": 1000,
  "telemetry_disabled": false,
  "alert_callback_url": "",
  "alert_consecutive_failure_count": 20,
  "alert_auto_disable_destination": true,
  "idgen_type": "uuidv4",
  "idgen_event_prefix": "",
}
```

Depending on the configured MQ, it will add the following fields:

**RabbitMQ:**
```json
{
  "rabbitmq_url": "amqp://***:***@rabbitmq:5672",
  "rabbitmq_exchange": "outpost",
  "rabbitmq_delivery_queue": "outpost-delivery",
  "rabbitmq_log_queue": "outpost-log"
}
```

**AWS SQS:**
```json
{
  "aws_access_key_configured": true,
  "aws_secret_key_configured": true,
  "aws_region": "us-east-1",
  "aws_delivery_queue": "queue-name"
  "aws_log_queue": "log-queue-name"
}
```

**GCP Pub/Sub:**
```json
{
  "gcp_credentials_configured": true,
  "gcp_project_id": "project_id"
  "gcp_delivery_topic": "topic",
  "gcp_delivery_subscription": "subscription-name",
  "gcp_log_topic": "log-topic",
  "gcp_log_subscription": "log-subscription-name"
}
```

**Azure Service Bus:**
```json
{
  "azure_connection_string_configured": true,
  "azure_delivery_topic": "topic",
  "azure_delivery_subscription": "subscription-name",
  "azure_log_topic": "log-topic",
  "azure_log_subscription": "log-subscription-name"
}
```

### DIFF
--- a/internal/config/logging.go
+++ b/internal/config/logging.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"fmt"
 	"strings"
 
 	"go.uber.org/zap"


alexbouchardd:
Need to be careful not to leak env secret values in the log!

fmvilas:
Yeah, that's why I'm masking URLs (in case they contain secrets). Also, I'm logging things like `aws_secret_key_configured: true` instead of logging the secret itself. Did you spot any other place?

leggetter:
We sanitize the API logging based on config that tells us the values that are sensitive. https://github.com/hookdeck/outpost/blob/main/internal/services/api/sanitizer.go

fmvilas:
> My biggest concern about this issue from the beginning was how to prevent drift as we continue building Outpost. Or we effectively have to add to our mental model (and probably should extend the contributing guide related to config?) that we need to account for this log whenever we add new configuration. Not a dealbreaker but something to consider.

Fair point. I'm going to address it with docs for now.

> One other idea that's a bit simpler is for each "section" of the config to have its own logging logic. For example, c.MQsConfig.ToLogFields(). Not 100% sure if that's better, just an idea.

Yeah, I feel you. I kinda did something similar by moving all these logs to `c.LogConfigurationSummary()`. I think making each section have its own logging logic feels more organized but can also feel scattered. I'm also not sure what would be best to be honest. Happy to go either way.

> We sanitize the API logging based on config that tells us the values that are sensitive. https://github.com/hookdeck/outpost/blob/main/internal/services/api/sanitizer.go

Thanks, @leggetter. Having a look to see if I can reuse some stuff. Although, these parts of the code are in two very different domains, if I'm understanding the codebase correctly. So I would hesitate from reusing directly. I could consider using a shared functionality in both. Perhaps a function inside "util".

fmvilas:
@alexluong Added docs to the Contribution Guide. I like your solution at #480 much better but I think it can be done in later on if it becomes a pain to maintain.

@leggetter I considered reusing some code but even though they do similar stuff, it feels like completely unrelated domains. I think creating an abstraction layer on top of both could actually be harder to maintain so I'm going to leave it as is.

PR #572, opened 2025-11-21
update zudoku

mosch:
Hey there, it's been a while since your Zudoku got updated. 
We just shipped a new version with plenty of updates & improvements. I would love if the Outpost documentation would be using the latest version!


leggetter:
Thanks for the PR @mosch. Looks like you and @dan-lee have been busy 🙌 

@fmvilas is leading Outpost, so this may also be an opportunity to familiarize himself with Zudoku.

PR #574, opened 2025-11-27
chore: portal UI enhancements

alexluong:
- Make sure the CTA button is above the fold

CleanShot 2025-11-27 at 19 27 31

- Improve "Select All" UX for topic picker

https://github.com/user-attachments/assets/002e78e5-626a-4509-887a-4ab64b9a2b2a

(small note but I don't know if you noticed there's a small bug around checkbox checked state going from indeterminate -> checked)

### DIFF
--- a/internal/portal/src/common/TopicPicker/TopicPicker.scss
+++ b/internal/portal/src/common/TopicPicker/TopicPicker.scss
@@ -4,6 +4,7 @@
   display: flex;
   flex-direction: column;
   overflow: hidden;
+  max-height: 100%;
 
   &__header {
     border-bottom: 1px solid var(--colors-outline-neutral);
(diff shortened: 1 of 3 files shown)


alexbouchardd:
The scroll area for event destinations is not contained. Instead of distinct cards I think we need to move to using a single big scrollable card when rows for each destination type

alexluong:
@alexbouchardd do you have screenshots or something? Not super clear to me what you mean just yet. I tried testing out the destination creation flow and thought it was good but yeah overall it was a bit tricky figuring out the scrollable section

alexbouchardd:
I'll handle!

alexbouchardd:
@alexluong Pushed fix
Screenshot 2025-11-27 at 8 55 08 PM

PR #577, opened 2025-12-02
feat: chlogstore supports multi-deployment settings

alexluong:
# Deployment ID Support for ClickHouse Log Store

## Summary

Adds deployment ID support to ClickHouse log store for multi-tenant isolation. Each deployment gets its own `{deploymentID}_event_log` table.

## Approach: Separate Tables per Deployment

Each deployment gets its own table (`{deploymentID}_event_log`) and migration tracking table (`{deploymentID}_schema_migrations`).

### Alternatives Considered

| Approach | Pros | Cons |
|----------|------|------|
| **Separate tables (chosen)** | Complete isolation, independent schema migrations, easy to manage/drop individual deployments | More tables to manage |
| **Partitioning by deployment_id** | Single table, built-in ClickHouse feature | Too many partitions (already partitioned by day), all deployments share same schema version |
| **deployment_id column in ORDER BY** | Simplest to operate (single table) | No per-deployment retention policy, query performance affected by total volume across all deployments |

**Note:** The deployment_id column approach is a strong alternative worth considering for simpler operational overhead. We chose separate tables primarily for per-deployment retention policies and query performance isolation.

### Why Separate Tables?

1. **Schema independence** - Each deployment can run different migration versions
2. **Failure isolation** - Issues in one deployment's table don't affect others
3. **Simple lifecycle management** - Easy to provision or tear down a deployment
4. **Flexible partitioning** - Each deployment can have different partitioning configuration and retention policies
5. **No query overhead** - No need to filter by deployment_id on every query

### DIFF
--- a/internal/logstore/chlogstore/chlogstore.go
+++ b/internal/logstore/chlogstore/chlogstore.go
@@ -20,12 +20,12 @@ type logStoreImpl struct {
 var _ driver.LogStore = (*logStoreImpl)(nil)
 
 // NewLogStore creates a new ClickHouse log store.
-// If deploymentID is provided, it uses a deployment-specific table (event_log_{deploymentID}).
+// If deploymentID is provided, it uses a deployment-specific table ({deploymentID}_event_log).
 // Otherwise, it uses the default "event_log" table.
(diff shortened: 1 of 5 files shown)


alexluong:
replaced by #645

PR #587, opened 2025-12-10
fix: Fix configuration documentation errors for `DESTINATIONS_WEBHOOK_HEADER_PREFIX` and `PORTAL_REFERER_URL`

alexbouchardd:


leggetter:
@alexbouchardd there is https://github.com/hookdeck/outpost/tree/main/cmd/configdocsgen that I had previously used to ensure we update both the env vars and yaml from attributes in the source code. Maybe it's out of date or obsolete? @alexluong

alexluong:
added a few commits to clean up the PR, merging now

@leggetter yes we're still using that script to keep things in sync

PR #588, opened 2025-12-10
chore: Remove timestamps from default signature template

alexbouchardd:
Including the timestamp into the default signature was an oversight.

There's a valid debate around the relevance of timestamp in signatures but our position is that it's security theater as does not effecively protect agaisnt replay attacks (given the need for a validity window). It also leads to more signature verification complexity which in turns results into less users verifying the signatures ultimately leading to less secure systems.

Outpost administrators can still decide to inclue the timestamp is they wish, it's simply not longer the default signature format.

This is a breaking change and to maintain current signature format configs need to be updated with:

```
DESTINATIONS_WEBHOOK_SIGNATURE_CONTENT_TEMPLATE={{.Timestamp.Unix}}.{{.Body}}
DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_TEMPLATE=t={{.Timestamp.Unix}},v0={{.Signatures | join ","}}
```

### DIFF
--- a/docs/pages/destinations/webhook.mdx
+++ b/docs/pages/destinations/webhook.mdx
@@ -172,8 +172,8 @@ Operators can customize webhook behavior via environment variables.
 | -------------------- | ------- | ----------- |
 | `DESTINATIONS_WEBHOOK_SIGNATURE_ALGORITHM` | `hmac-sha256` | Signature algorithm |
 | `DESTINATIONS_WEBHOOK_SIGNATURE_ENCODING` | `hex` | Signature encoding (`hex` or `base64`) |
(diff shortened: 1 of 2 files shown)


alexluong (commented):
let's include this in the upcoming release with CH & Log API given it's a pretty big breaking change

also will need some changes in the tests I think but I can take it from here

alexluong:
TODO: make sure the generated config md shows the right value for `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_TEMPLATE=t={{.Timestamp.Unix}},v0={{.Signatures | join ","}}`

Issue #586, opened 2025-12-09
RFC: Destination Filter


alexluong:
duplicates #492 

# Feature: Event Filtering for Destinations

## Summary

Add an optional filter to destinations, enabling event routing based on event content—not just topic.

**Current Flow:**
```
Publish Event → Match by Topic → Enqueue to DeliveryMQ
```

**Proposed Flow:**
```
Publish Event → Match by Topic → Filter → Enqueue to DeliveryMQ
```

## Motivation

Topic-based matching provides coarse-grained routing (e.g., `order.created`, `user.updated`), but many use cases require finer control. Filters let destinations selectively receive events based on their content.

### Destination Schema Change

Add an optional `filter` field to the `Destination` model. When set, events must match the filter to be delivered to that destination.

### Event Schema for Filtering

The filter operates on the following event structure:

```json
{
  "id": "evt_123",
  "topic": "order.created",
  "time": "2025-01-15T10:30:00Z",
  "metadata": {
    "source": "mobile-app",
    "region": "us-east",
    "environment": "production"
  },
  "data": {
    "order_id": "ord_456",
    "amount": 1500,
    "status": "confirmed",
    "customer": {
      "tier": "premium"
    }
  }
}
```

| Field | Type | Description |
|-------|------|-------------|
| `id` | string | Unique event identifier |
| `topic` | string | Event topic for routing |
| `time` | string (ISO 8601) | Event timestamp |
| `metadata` | object | Key-value pairs for routing context (source, region, environment, etc.) |
| `data` | object | Event payload |

### Filter Syntax

The filter syntax follows a JSON matching pattern similar to [Hookdeck Filters](https://hookdeck.com/docs/filters), implemented via the [simple-json-match](https://github.com/hookdeck/simple-json-match) library (to be ported to Go).

#### Supported Operators

| Operator | Types | Description |
|----------|-------|-------------|
| `$eq` | any | Deep equality check |
| `$neq` | any | Deep inequality check |
| `$gt` | number, string | Greater than |
| `$gte` | number, string | Greater than or equal |
| `$lt` | number, string | Less than |
| `$lte` | number, string | Less than or equal |
| `$in` | array, string | Substring or array membership |
| `$nin` | array, string | Negated membership |
| `$startsWith` | string | Prefix matching |
| `$endsWith` | string | Suffix matching |
| `$exist` | boolean | Check field presence |
| `$or` | array | Any condition matches |
| `$and` | array | All conditions match |
[...]

PR #594, opened 2025-12-11
feat: implement destination filter

alexluong:
implements #586

# Destination Filter

Adds support for event filtering on destinations, enabling event routing based on event content—not just topic.

## Features

- **Filter Field**: New optional `filter` field on destinations allows filtering events based on JSON matching patterns
- **simplejsonmatch Library**: Go port of [hookdeck/simple-json-match](https://github.com/hookdeck/simple-json-match) with 1-to-1 mapping of test cases from the original JS implementation. **Note:** The `$ref` operator is not supported in this implementation.
- **Portal UI**: JSON editor for managing filters in destination settings and create destination flow.

## UI

https://github.com/user-attachments/assets/5fc8b71c-270b-4451-9cdf-d647d8df5851


alexbouchardd:
Should this be opt-in the same way topics are? 🤔

alexluong:
> Should this be opt-in the same way topics are? 🤔

Not 100% sure what you mean, I think it is opt-in. As in `destination.filter` is optional, if nothing is provided then it's automatically matched based on topic only.

alexbouchardd:
Some `ENABLE_DESTINATION_FILTERS` env variable is what I meant. Are we confident everyone would like that supported? 🤔

alexluong:
ah I see. Hmmm not sure, I don't see a strong reason why someone would prefer not to have this feature. Maybe performance concern?

If someone is opinionated enough, maybe they already provide their own UI and can choose not to use this feature? Maybe we can support a flag to disable this in the portal only?

I just think disabling this feature as a whole may be a bit complicated because of the Destination schema (Destination API).

alexbouchardd:
Yeah agree really that's a portal config if anything since if you implement your own UI then you get to choose that way.

The problem if we don't give control over it is that a user may deploy the portal and later want to build their own UI or migrate to another system and be stuck with that backward compatibility. Given it does create some amount of "lock-in" with Outpost, it should be something you can decide to expose to your users or not.

`PORTAL_DISABLE_FILTER_CONFIGURATION`
`PORTAL_DISABLE_WEBHOOK_DESTINATION_CUSTOM_HEADERS`

Does that make sense?

alexluong:
yup, all clear then. Please continue finalizing the UI changes the way you see fit. I'll take care of the portal flag for these 2 features. Will merge and release if you approve the PR

docs/apis/openapi.yaml
```diff
@@ -99,6 +99,24 @@ components:
       description: '"*" or an array of enabled topics.'
       example: "*"
 
+    Filter:
+      type: object
+      nullable: true
+      additionalProperties: true
+      description: |
```
alexbouchardd, line 106: The `{}` doesn't make sense to me, a user should pass `null` to unset

alexluong: hey so quick note, i'll merge to a temp `v0.10.0` branch now but we can discuss this further

---

good catch, I do agree that `{}` is not ideal. The reason tho is that the way we parse incoming request, in Golang `null` and `undefined` are the same -> `nil`, and for PATCH we treat nil as no update. We can still support `null` but it requires a bit more work which we can do, but I think this is an acceptable tradeoff for now. Do let me know tho. I agree, I don't love this DX and spent quite a bit of time thinking on this exact question

PR #600, opened 2025-12-16
fix: treat webhook connection errors as delivery errors (retry) instead of system errors 

### LINKED ISSUE #571: Clarification on DLQ usage and handling of user-caused delivery failures in SQS setups
We’re running Outpost with SQS and noticed some DLQ behavior we’d like to understand better.

### Context

Our SQS DLQ regularly receives messages that correspond to **normal user-caused delivery failures** (e.g., invalid webhook URLs and endpoints returning 4xx). These aren’t system issues, but more like expected delivery failures that Outpost retries and tracks.

However, because the worker never reaches a “successful processing” state for these events, SQS eventually moves them to the DLQ.

### The question

We think DLQ should represent **pipeline-level failures** (parsing errors, crashes, internal outages), while user-misconfigurations should be treated as handled delivery failures within Outpost.

So we want to understand:

1. Is it expected that user-level delivery failures end up in the DLQ?
2. Should these failures be acknowledged as “processed” from the queue’s perspective, even if the webhook call failed?
3. Is there a recommended approach to avoid mixing user misconfigurations with actual Outpost/system failures in the DLQ?
4. Would enabling failure alerts (`ALERT_CALLBACK_URL`) change how messages are acknowledged?

### Goal

We would like DLQ to signal **real processing issues**, not just tenants misconfiguring their webhook destinations. Any clarification or guidance would be greatly appreciated!

Thanks for the great project!

alexluong:
fixes #571 

### Problem

Connection errors (DNS failures, connection refused, etc.) were being sent to DLQ instead of being retried. This happened because these errors were classified as "pre-delivery errors" rather than "delivery errors".

### Solution

Network errors are now treated as delivery errors, which means they will be acknowledged and scheduled for retry instead of being nacked and sent to DLQ.

### Error Classification

| Code | Cause | Type | Behavior |
|------|-------|------|----------|
| `dns_error` | Domain doesn't exist | Delivery Error | ack + retry |
| `connection_refused` | Server not running | Delivery Error | ack + retry |
| `connection_reset` | Connection dropped | Delivery Error | ack + retry |
| `network_unreachable` | Network unavailable | Delivery Error | ack + retry |
| `timeout` | I/O timeout or deadline | Delivery Error | ack + retry |
| `tls_error` | Certificate/TLS failure | Delivery Error | ack + retry |
| `redirect_error` | Too many redirects | Delivery Error | ack + retry |
| `network_error` | Other network issues | Delivery Error | ack + retry |
| `canceled` | Context canceled (shutdown) | System Error | nack + requeue |

The fix applies to both `webhook` and `webhook_standard` destination types.

### DIFF
--- a/internal/destregistry/providers/destazureservicebus/destazureservicebus.go
+++ b/internal/destregistry/providers/destazureservicebus/destazureservicebus.go
@@ -159,28 +159,28 @@ func (p *AzureServiceBusPublisher) Publish(ctx context.Context, event *models.Ev
 	sender, err := p.ensureSender()
 	if err != nil {
 		return &destregistry.Delivery{
-			Status: "failed",
-			Code:   "ERR",
-			Response: map[string]interface{}{
-				"error": err.Error(),
-			},
-		}, destregistry.NewErrDestinationPublishAttempt(err, "azure_servicebus", map[string]interface{}{
-			"error":   "sender_failed",
-			"message": err.Error(),
-		})
+				Status: "failed",
+				Code:   "ERR",
+				Response: map[string]interface{}{
+					"error": err.Error(),
+				},
+			}, destregistry.NewErrDestinationPublishAttempt(err, "azure_servicebus", map[string]interface{}{
+				"error":   "sender_failed",
+				"message": err.Error(),
+			})
 	}
 
 	if err := sender.SendMessage(ctx, message, nil); err != nil {
 		return &destregistry.Delivery{
-			Status: "failed",
-			Code:   "ERR",
-			Response: map[string]interface{}{
-				"error": err.Error(),
-			},
-		}, destregistry.NewErrDestinationPublishAttempt(err, "azure_servicebus", map[string]interface{}{
-			"error":   "send_failed",
-			"message": err.Error(),
-		})
+				Status: "failed",
+				Code:   "ERR",
+				Response: map[string]interface{}{
+					"error": err.Error(),
+				},
+			}, destregistry.NewErrDestinationPublishAttempt(err, "azure_servicebus", map[string]interface{}{
+				"error":   "send_failed",
+				"message": err.Error(),
+			})
 	}
 
 	return &destregistry.Delivery{
--- a/internal/destregistry/providers/destrabbitmq/destrabbitmq.go
+++ b/internal/destregistry/providers/destrabbitmq/destrabbitmq.go
@@ -140,15 +140,15 @@ func (p *RabbitMQPublisher) Publish(ctx context.Context, event *models.Event) (*
 
 	if err := p.ensureConnection(ctx); err != nil {
 		return &destregistry.Delivery{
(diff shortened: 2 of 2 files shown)


alexbouchardd:
>The fix applies to both webhook and webhook_standard destination types.

Can you clarify what the behavior is for say RabbitMQ if the queue is unavailable / has incorrect credentials? That should also retry. Anything that is outside of Outpost control really, should retry and not go to DLQ

alexluong:
> Can you clarify what the behavior is for say RabbitMQ if the queue is unavailable / has incorrect credentials?

I can look into RabbitMQ later. What do you think if we move forward with this PR and have RabbitMQ as a follow-up?

As a whole we need to be a bit more intentional about the error handling for every destination type. This PR focuses on webhook for now.

If you prefer we handle everything in this PR, happy to do so as well.

alexbouchardd:
I'm just using RabbitMQ as an example of all other destination types here but no, I think that's fundamental design for the error handling where destination specific config shouldn't cause message to go to DQL and that's a generalized implementation rather then destination per destination

alexluong:
Here's the delivery logic

```
deliverymq message handler:
  pre_delivery_err, delivery_err = deliver(destination, event)
  if pre_delivery_err -> nack
  if delivery_err -> schedule retry
  ack
```

It just happens that in webhook, the HTTP request is splitted into 2 steps. I am also just now realized that this PR basically will ignore cases where the system loses network, or if its DNS resolver has issue. That's why I said it's not that simple, at least from what I'm seeing.

For example, this is the step in RabbitMQ that would qualify as PreDelivery

CleanShot 2025-12-16 at 22 46 46

This could be because auth issue. Or RabbitMQ destination is down. But what if it's because some system issue? How can we differentiate?

alexbouchardd:
If we don't have a connection to rabbitMQ (or any destination for that matter) it's a delivery error, not "pre delivery". It's not "Outpost" fault. The only things that would outpost fault would be unexpected data in the message, errors fetch the destination config (Redis calls), parsing or application logic errors

alexbouchardd:
Ultimately it's about who's responsible for the issue, the operator or the tenant

alexluong:
I get that, I guess it doesn't quite click for me what's the simplest way to differentiate.

For example, when we see an error as "DNS resolution" or "connection reset" error or "generic network error", is that 100% of the time destination error?

alexbouchardd:
Yes it is, or least we treat it that way since its transient

alexluong:
okay noted, let me noodle on this and apply this idea to the other destinations.

For now, I'd like to move forward with v0.10.0 without this PR, then focus on the CH implementation, then I'll get back to this. Do let me know if you prefer I prioritize this instead.

alexbouchardd:
Yep make sense to me but this needs to resolved and bulletproof the 1.0

alexbouchardd (approved):

PR #602, opened 2025-12-16
refactor: portal sidebar

alexluong:
# Sidebar Refactor

## Problem

Previously, `ConfigurationModal` and `FilterSyntaxModal` each created their own portal with `id="sidebar"`. When both were open simultaneously, they would overlap and render unreadable content.

CleanShot 2025-12-16 at 21 48 39

## Solution

Introduced a shared `Sidebar` context that manages a single sidebar instance. Only one sidebar can be visible at a time - opening a new one replaces any existing sidebar.

## API

```tsx
import { useSidebar } from "../Sidebar/Sidebar";

const sidebar = useSidebar();

// Toggle - opens if closed, closes if same ID is already open
sidebar.toggle("filter-syntax", <FilterSyntaxGuide />);

// Always opens (replaces current sidebar if different ID)
sidebar.open("filter-syntax", <FilterSyntaxGuide />);

// Closes only if that ID is currently open
sidebar.close("filter-syntax");
```

## Usage Example

```tsx
const MyComponent = () => {
  const sidebar = useSidebar();

  return (
    <Button
      onClick={() =>
        sidebar.toggle("filter-syntax", <FilterSyntaxGuide />)
      }
    >
      Filter Syntax Guide
    </Button>
  );
};
```


internal/portal/src/common/FilterSyntaxGuide/FilterSyntaxGuide.tsx
```diff
@@ -1,8 +1,3 @@
-import { useEffect, useState } from "react";
-import { createPortal } from "react-dom";
-import { CollapseIcon } from "../Icons";
-import Button from "../Button/Button";
-import "../ConfigurationModal/ConfigurationModal.scss";
 import Markdown from "react-markdown";
 
 const FILTER_SYNTAX_GUIDE = `# Filter Syntax Guide
```
alexbouchardd, line 3: @alexluong Just now realizing that users may want to render this in their own UI just like the config instructions. What you do reckon would be the best way to expose that via the API? 🤔 Doesn't block this PR

alexluong: ah I assume we're talking about Filter syntax. Maybe we can expose an endpoint and move this to the server as md like instruction.

GET /instructions/filter
GET /instructions/destination/webhook
...

We don't need the other ones for now since they're already included in the destination types endpoint.

alexbouchardd: We don't have a `/instructions` api root though, the intrusctions are within the destination type schemas

alexluong: yes, I was suggesting a convention if we want to go this route?

"We don't need the other ones _(other than filter)_ for now since they're already included in the destination types endpoint."

alexbouchardd: Let's ignore and merge as is for now

PR #614, opened 2026-01-08
fix: add new az params and update to v11

Fraaankes:
Fixes a couple of warnings when trying out the Azure example

https://learn.microsoft.com/en-us/cli/azure/upcoming-breaking-changes?view=azure-cli-latest#redisenterprise-create
> The argument '--public-network-access' will become required in next breaking change release(2.79.0) scheduled for Nov 2025.
The default value of '--access-keys-auth' will be changed to 'Disabled' from 'Enabled' in next breaking change release(2.79.0) scheduled for Nov 2025.

and updates to the latest docker image


leggetter:
Thanks @Fraaankes 🙌

PR #617, opened 2026-01-13
feat: add /tenants prefix to tenant-scoped API routes

alexluong:
Adds `/tenants` prefix to all tenant-scoped routes

## Changes

- New canonical paths: `/api/v1/tenants/:tenantID/*`
- Old paths (`/api/v1/:tenantID/*`) still work but return `Deprecation` and `Link` headers
- Updated OpenAPI spec to reflect new paths


alexbouchardd:
> Old paths (/api/v1/:tenantID/*) still work but return Deprecation and Link headers

I would remove this and force the upgrade

PR #652, opened 2026-01-20
feat: clickhouse tls support + fix bugs + fix portal support new schema

alexluong:


alexbouchardd:
I believe clickhouse also needs to be added to the docs? 🤔

PR #653, opened 2026-01-21
chore: remove delivery event concept

alexluong:
## Remove DeliveryEvent concept

This PR replaces the monolithic `DeliveryEvent` type with purpose-specific types that better represent the different stages of event delivery.

### New Types

- **`DeliveryTask`** - Message type for delivery processing queues. Flows from `publishmq` → `deliverymq` and from retry → `deliverymq`. Contains the `Event`, `DestinationID`, `Attempt`, and `Manual` flag. Provides `IdempotencyKey()` for deduplication.

- **`RetryTask`** - Message type for scheduling retries. Contains only the metadata needed to re-fetch the event and re-attempt delivery (`EventID`, `TenantID`, `DestinationID`, `Attempt`). Converts to `DeliveryTask` when the retry fires.

- **`LogEntry`** - Message type for the log queue. Contains both `Event` and `Delivery` for persistence to the logstore.

- **`DeliveryRecord`** - Query result type returned by logstore. Contains `Delivery` with optional `Event` population for API responses.

### Logstore Interface Changes

- `InsertManyDeliveryEvent` → `InsertMany(events, deliveries)`
- `ListDeliveryEvent` → `ListDelivery` (returns `DeliveryRecord`)
- `RetrieveDeliveryEvent` → `RetrieveDelivery` (returns `DeliveryRecord`)

### Other Changes

- Removes `delivery_event_id` column from database schema
- Removes legacy `/delivery-events` API endpoints
- Idempotency now uses `event_id:destination_id` format (with `:manual` suffix for manual retries)

### DIFF
--- a/internal/portal/src/scenes/Destination/Destination.tsx
+++ b/internal/portal/src/scenes/Destination/Destination.tsx
@@ -24,7 +24,7 @@ interface Tab {
 const tabs: Tab[] = [
   { label: "Overview", path: "" },
   { label: "Settings", path: "/settings" },
-  { label: "Event Deliveries", path: "/attempts" },
+  { label: "Event Deliveries", path: "/deliveries" },
 ];
 
 const Destination = () => {
@@ -131,7 +131,7 @@ const Destination = () => {
               }
             />
             <Route
-              path="/attempts/*"
+              path="/deliveries/*"
               element={<AttemptRoutes destination={destination} />}
             />
             <Route
--- a/internal/portal/src/scenes/Destination/Events/Attempts.tsx
+++ b/internal/portal/src/scenes/Destination/Events/Attempts.tsx
@@ -131,7 +131,7 @@ const Attempts: React.FC<AttemptsProps> = ({
     <div className="destination-attempts">
       <div className="destination-attempts__header">
         <h2 className="destination-attempts__header-title title-l">
-          Attempts{" "}
+          Event Deliveries{" "}
           <Badge text={attemptsList?.models.length ?? 0} size="s" />
         </h2>
         <div className="destination-attempts__header-filters">
@@ -396,7 +396,7 @@ export const AttemptRoutes = ({ destination }: { destination: any }) => {
   const navigateAttempt = useCallback(
     (path: string, state?: any) => {
       navigate(
-        `/destinations/${destination.id}/attempts${path}?${urlSearchParams}`,
+        `/destinations/${destination.id}/deliveries${path}?${urlSearchParams}`,
         { state },
       );
     },


internal/apirouter/router.go
alexbouchardd, line 1: Attempts belong to a destination, unclear to me why we removed `/tenants/:tenant_id,/destinations/:dest_id/attempts`

internal/deliverymq/retry.go
```diff
@@ -32,68 +59,99 @@ func NewRetryScheduler(deliverymq *DeliveryMQ, redisConfig *redis.RedisConfig, d
...
+		if err != nil {
+			// Transient error (DB connection issue, etc) - return error so scheduler retries
+			if logger != nil {
+				logger.Ctx(ctx).Error("failed to fetch event for retry",
+					zap.Error(err),
+					zap.String("event_id", retryTask.EventID),
+					zap.String("tenant_id", retryTask.TenantID),
+					zap.String("destination_id", retryTask.DestinationID))
+			}
 			return err
```
alexbouchardd, line 92: What happens in this case? The message isn't removed from Redis and then what?

internal/models/event.go
```diff
@@ -67,75 +66,105 @@ func (e *Event) ToMessage() (*mqs.Message, error) {
...
+	}
+	return &mqs.Message{Body: data}, nil
+}
+
+type Attempt struct {
+	ID            string                 `json:"id"`
+	TenantID      string                 `json:"tenant_id"`
+	EventID       string                 `json:"event_id"`
+	DestinationID string                 `json:"destination_id"`
+	AttemptNumber int                    `json:"attempt"`
```
alexbouchardd: That seems a bit odd label, what about just `Number`?

internal/portal/src/common/RetryAttemptButton/RetryAttemptButton.tsx
```diff
@@ -24,12 +24,12 @@ const RetryDeliveryButton: React.FC<RetryDeliveryButtonProps> = ({
   const apiClient = useContext(ApiContext);
   const [retrying, setRetrying] = useState<boolean>(false);
 
-  const retryDelivery = useCallback(
+  const retryAttempt = useCallback(
```
alexbouchardd: I'm hesitant on that terminology, you aren't really retrying an attempt, you retry a delivery with generates and attempt. Thoughts?

internal/portal/src/scenes/Destination/Destination.tsx
```diff
@@ -14,19 +14,17 @@ import {
...
   label: string;
   path: string;
 }
 
-// Define available tabs
 const tabs: Tab[] = [
   { label: "Overview", path: "" },
   { label: "Settings", path: "/settings" },
-  { label: "Deliveries", path: "/deliveries" },
+  { label: "Attempts", path: "/attempts" },
```
alexbouchardd: I don't think Attempts is the right term for the UI, end-users don't think in terms of Attempts. I would call it "Event deliveries" which shows a list of delivery attemtps. This is what Stripe does

internal/apirouter/router.go
alexluong: I was thinking that this is the same as `/tenants/:tenant_id/attempts?destination_id=dest_id`, so it's effectively syntactic sugar. I thought it was simpler to have only 1 way of doing things.

Not an issue to add this back, so let me know which you prefer.

internal/deliverymq/retry.go
alexluong: RSMQ supports visibility timeout, so it will be visible again for try later

internal/models/event.go
alexluong: will also include a migration to change  field `attempt` -> `number`, right?

FWIW Hookdeck also calls it `attempt_number` ([ref](https://hookdeck.com/docs/api#attempts))

CleanShot 2026-01-29 at 01 12 20

internal/deliverymq/retry.go
alexbouchardd: Right although that's difficult to observe & monitor. There's also effectively no DQL here. Okay for now but something to keep in mind.

internal/models/event.go
alexbouchardd: The json in the line commented on is just `attempt` though. I could also go with `attempt_number` I just think `attempt` is vague

internal/apirouter/router.go
alexbouchardd: I think from a REST principle attempts belong to a destination and so have a explicit `/tenants/:tenant_id,/destinations/:dest_id/attempts` is more idiomatic? 🤔

internal/models/event.go
alexluong: yes, I’ll change to “attempt_number” to be consistent with Hookdeck API

“number” is a tad vague

alexluong: i'm working on this attempt_number thing and i noticed some inconsistency & things to discuss. Updated the naming in this PR and created this issue #662 here

internal/apirouter/router.go
alexluong: added

GET /:tenantID/destinations/:destinationID/attempts
GET /:tenantID/destinations/:destinationID/attempts/:attemptID
POST /:tenantID/destinations/:destinationID/attempts/:attemptID/retry

internal/portal/src/common/RetryAttemptButton/RetryAttemptButton.tsx
alexluong: i changed this on the FE but I assume your feedback is more general. My concern here is how would API for 'retry delivery' work since we don't expose the concept of delivery?

so should we do 'retry event to a destination' as a requirement?

```
# instead of
POST /tenants/:tenant_id/attempts/:attempt_id/retry

# do something along this line
POST /tenants/:tenant_id/destinations/:destination_id/events/:event_id/retry
#or
POST /tenants/:tenant_id/events/:event_id/retry?destination_id=...
```

it just feels a bit awkward I think

internal/portal/src/scenes/Destination/Destination.tsx
alexluong: I changed to "Event Deliveries" but kept the path `/attempts`, or do you think `/deliveries` make more sense?

alexbouchardd: For the front end yes

internal/portal/src/common/RetryAttemptButton/RetryAttemptButton.tsx
alexbouchardd: This seems like the most idiomatic: `POST /tenants/:tenant_id/destinations/:destination_id/events/:event_id/retry`

PR #665, opened 2026-01-29
fix: use correct default header prefix for standard webhook mode

### LINKED ISSUE #664: Webhook mode standard delivers wrong header names
**Describe the bug**
After setting the env `DESTINATIONS_WEBHOOK_MODE=standard` i am observing that my webhook destination is receiving the following headers:

```
x-outpost-topic: order.completed
x-outpost-timestamp: 1769699475
x-outpost-signature: v1,dJtLfWX/Nx7JJ8MXtyucwceTrhzD+qRMHigT0xJ/xsE=
x-outpost-id: 4ba74706-4744-4054-8c6d-0d7fe604b356
``` 

Based on the v1 signature, it appears to be hashing according to the standards mode, but the header name prefix is supposed to be `webhook-` according to the documentation.
Note that I do not have `DESTINATIONS_WEBHOOK_HEADER_PREFIX` set, and I could probably fix it by setting it myself, but this seems like a bug.

**To Reproduce**
1. Set webhook mode to standard
2. Add a destination
3. Publish an event
4. Observe the header names received

**Expected behavior**
Received headers matches according too the documentation

> Prefix for metadata headers added to webhook requests. Defaults to 'x-outpost-' in 'default' mode and 'webhook-' in 'standard' mode.

**Screenshots**

**Additional context**
Tracing through the code a bit, I believe the problem comes from here https://github.com/hookdeck/outpost/blob/main/internal/destregistry/providers/default.go#L59 where the global default value, is overriding the standard default value from here https://github.com/hookdeck/outpost/blob/main/internal/destregistry/providers/destwebhookstandard/destwebhookstandard.go#L123 when the opts are applied

alexluong:
fixes #664

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -176,7 +176,6 @@ func (c *Config) InitDefaults() {
 	c.Destinations = DestinationsConfig{
 		MetadataPath: "config/outpost/destinations",
 		Webhook: DestinationWebhookConfig{
-			HeaderPrefix:             "x-outpost-",
 			SignatureContentTemplate: "{{.Body}}",
 			SignatureHeaderTemplate:  "v0={{.Signatures | join \",\"}}",
 			SignatureEncoding:        "hex",
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -71,7 +71,7 @@ func TestDefaultValues(t *testing.T) {
 	assert.Equal(t, "config/outpost/destinations", cfg.Destinations.MetadataPath)
 	assert.Equal(t, 10, cfg.LogBatchThresholdSeconds)
 	assert.Equal(t, 1000, cfg.LogBatchSize)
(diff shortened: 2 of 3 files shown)


alexluong:
> Confused by the changes here it seems like we are removed the default instead of replacing by x-webhook when in standard mode

right, so the issue was we're setting default in multiple places, i.e. in the config & also in the [`destwebhook`](https://github.com/hookdeck/outpost/blob/b269b785fe66888d152bfcec3d5352c1d0f25fbd/internal/destregistry/providers/destwebhook/destwebhook.go#L192C3-L192C15) & [`destwebhookstandard`](https://github.com/hookdeck/outpost/blob/b269b785fe66888d152bfcec3d5352c1d0f25fbd/internal/destregistry/providers/destwebhookstandard/destwebhookstandard.go#L123) itself. So by removing the default in config, we're allowing each destination default to be used.

CleanShot 2026-01-31 at 00 07 15

CleanShot 2026-01-31 at 00 06 57

PR #667, opened 2026-01-29
docs: upgrade v0.13

alexluong:


docs/pages/guides/upgrade-v0.13.mdx
```diff
@@ -0,0 +1,183 @@
...
+The "delivery" concept in the API has been renamed to "attempt." An attempt represents a single delivery attempt of an event to a destination — the same concept, just a clearer name.
+
+### API Routes
+
+| v0.12 Route | v0.13 Route |
+| --- | --- |
+| `GET /tenants/:tenant_id/events/:event_id/deliveries` | **Removed** — use `GET /tenants/:tenant_id/attempts?event_id=:event_id` |
+| `GET /tenants/:tenant_id/destinations/:destination_id/events` | **Removed** — use `GET /tenants/:tenant_id/attempts?destination_id=:destination_id` |
+| `GET /tenants/:tenant_id/destinations/:destination_id/events/:event_id` | **Removed** |
+| `POST /tenants/:tenant_id/destinations/:destination_id/events/:event_id/retry` | `POST /tenants/:tenant_id/attempts/:attempt_id/retry` |
```
alexbouchardd: Needs review that's no longer correct

docs/pages/guides/upgrade-v0.13.mdx
```diff
@@ -0,0 +1,183 @@
...
+| `POST /tenants/:tenant_id/attempts/:attempt_id/retry` | Retry a specific attempt |
+
+### Attempt Schema
+
+The new `Attempt` resource looks like this:
+
+```json
+{
+  "id": "atm_123",
+  "status": "success",
```
alexbouchardd, line 28: event_id and destination_id missing?

docs/pages/guides/upgrade-v0.13.mdx
```diff
@@ -0,0 +1,183 @@
...
+   - [ ] Append delimiter to all custom `IDGEN_*_PREFIX` config values (e.g., `evt` → `evt_`)
+   - [ ] Update response parsing for the new pagination envelope (`models`/`pagination`)
+   - [ ] Update time filter params from `start`/`end` to `time[gte]`/`time[lte]`
+   - [ ] Update error response handling if parsing `data` field
+   - [ ] Back up Redis data
+
+2. **During upgrade:**
+   - [ ] Stop Outpost
+   - [ ] Run migrations: `migrate apply --all`
+   - [ ] Verify migrations completed: `migrate verify`
```
alexbouchardd: Given the tables have changes won't that lead to data lost?

alexluong: The migrate command is only for Redis. For Postgres it's automatic, and also we used `ALTER TABLE` command.

With that said, we should remove this step. I also probably should QA the migration step from v0.12 -> v0.13 for Postgres users.

alexbouchardd: Okay great ALTER works and since clickhouse is net new it's not an issue 👍

docs/pages/references/configuration.mdx
```diff
@@ -57,7 +57,7 @@ Global configurations are provided through env variables or a YAML file. ConfigM
 | `DESTINATIONS_WEBHOOK_DISABLE_DEFAULT_SIGNATURE_HEADER` | If true, disables adding the default 'X-Outpost-Signature' header to webhook requests. Only applies to 'default' mode. | `false` | No |
 | `DESTINATIONS_WEBHOOK_DISABLE_DEFAULT_TIMESTAMP_HEADER` | If true, disables adding the default 'X-Outpost-Timestamp' header to webhook requests. Only applies to 'default' mode. | `false` | No |
 | `DESTINATIONS_WEBHOOK_DISABLE_DEFAULT_TOPIC_HEADER` | If true, disables adding the default 'X-Outpost-Topic' header to webhook requests. Only applies to 'default' mode. | `false` | No |
-| `DESTINATIONS_WEBHOOK_HEADER_PREFIX` | Prefix for metadata headers added to webhook requests. Defaults to 'x-outpost-' in 'default' mode and 'webhook-' in 'standard' mode. | `x-outpost-` | No |
+| `DESTINATIONS_WEBHOOK_HEADER_PREFIX` | Prefix for metadata headers added to webhook requests. Defaults to 'x-outpost-' in 'default' mode and 'webhook-' in 'standard' mode. | `nil` | No |
```
alexbouchardd, line 60: Oversight? We still default to `x-outpost-`

alexluong: so this is from a config auto-gen thing. We briefly discussed this before, but the idea is there's a config default and then there's feature (destination) default.

destwebhook defaults to `x-outpost-`
destwebhookstandard defaults to `webhook-`

if config default to a value, then it will override the destination default. So if we want to have a config default, then each destination's default becomes meaningless. So we need to then parse the mode in config to set the right default before passing that to the destination.

anyway, that's to explain why the default here is `nil` which is correct

alexbouchardd: Right okay fair enough

PR #674, opened 2026-02-04
fix: enforce validation for empty custom_headers in webhook configurations

alexbouchardd:
- Updated tests to ensure that an error is returned when custom_headers is empty or invalid.
- Modified the WebhookDestination and StandardWebhookDestination to return validation errors for empty custom_headers.
- Adjusted frontend logic to handle empty custom_headers appropriately, ensuring backward compatibility while enforcing validation rules.

### DIFF
--- a/internal/destregistry/providers/destwebhookstandard/destwebhookstandard_config_test.go
+++ b/internal/destregistry/providers/destwebhookstandard/destwebhookstandard_config_test.go
@@ -52,7 +52,7 @@ func TestStandardWebhookDestination_CustomHeadersConfig(t *testing.T) {
 		var validationErr *destregistry.ErrDestinationValidation
 		assert.ErrorAs(t, err, &validationErr)
 		assert.Equal(t, "config.custom_headers", validationErr.Errors[0].Field)
-		assert.Equal(t, "required", validationErr.Errors[0].Type)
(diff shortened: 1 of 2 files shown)


internal/destregistry/providers/destwebhookstandard/destwebhookstandard_config_test.go
```diff
@@ -48,7 +48,11 @@ func TestStandardWebhookDestination_CustomHeadersConfig(t *testing.T) {
 		)
 
 		err := provider.Validate(context.Background(), &destination)
-		assert.NoError(t, err)
+		assert.Error(t, err)
+		var validationErr *destregistry.ErrDestinationValidation
+		assert.ErrorAs(t, err, &validationErr)
+		assert.Equal(t, "config.custom_headers", validationErr.Errors[0].Field)
+		assert.Equal(t, "required", validationErr.Errors[0].Type)
```
alexluong: hey the test suite is failing because the assertion here is incorrect. Based on the error you're throwing, it should be "invalid" instead of "required"

Image

alexbouchardd: Indeed missed that one!

PR #701, opened 2026-02-11
fix: add PubSub retry policy and increase ack deadline

alexluong:
Increase default ack deadline from 10s to 60s to prevent premature redelivery when batch processing takes up to ~40s (10s batch delay + 30s InsertMany timeout).

Add RetryPolicy with exponential backoff (10s min, 600s max) on nacked messages to avoid tight retry loops when ClickHouse is unavailable.

Note: existing subscriptions need to be updated manually as the infra provisioner only applies settings on creation.

### DIFF
--- a/internal/mqinfra/gcppubsub.go
+++ b/internal/mqinfra/gcppubsub.go
@@ -12,6 +12,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	defaultMinRetryBackoffSeconds = 10
+	defaultMaxRetryBackoffSeconds = 120
+)
+
 type infraGCPPubSub struct {
 	cfg *MQInfraConfig
 }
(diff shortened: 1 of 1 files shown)


internal/mqinfra/gcppubsub.go
```diff
@@ -158,7 +158,7 @@ func (infra *infraGCPPubSub) Declare(ctx context.Context) error {
 	}
 
 	// Set visibility timeout (acknowledgement deadline)
-	ackDeadline := 10 // default 10 seconds
+	ackDeadline := 60 // default 60 seconds
```
alexluong, line 166: the thought process for 60s is

30s CH timeout
10s of max waiting in the batch

i'm setting 60s just to round it up, in a sense. Could be 45s too.

Also this is applicable for basically both deliverymq and logmq, do let me know if you think it makes sense to separate them somehow?

internal/mqinfra/gcppubsub.go
```diff
@@ -182,6 +182,10 @@ func (infra *infraGCPPubSub) Declare(ctx context.Context) error {
 				DeadLetterTopic:     dlqTopic.String(),
 				MaxDeliveryAttempts: maxDeliveryAttempts,
 			},
+			RetryPolicy: &pubsub.RetryPolicy{
+				MinimumBackoff: getRetryBackoff(infra.cfg.GCPPubSub.MinRetryBackoff, 10),
+				MaximumBackoff: getRetryBackoff(infra.cfg.GCPPubSub.MaxRetryBackoff, 600),
+			},
```
alexluong, line 193: i'm setting 10s min and 600s max, let me know if you think there's a different / better configuration for backoff logic

alexbouchardd (approved):

internal/mqinfra/gcppubsub.go
alexbouchardd: I think 600s is too long I'd shorten to ~120s

PR #706, opened 2026-02-14
fix: API handler fixes and OpenAPI spec audit

alexluong:
## API fixes

- **Publish: require `data` field** — was silently accepting events without payload
- **Destination update: reject type changes** — PATCH was allowing type to be changed, now returns 422 (sending same type still allowed for SDK compat)
- **List tenants: validate `limit` range** — enforce 1-100 bounds to match spec

## OpenAPI spec fixes

- Add `components/responses` for standard errors (400, 401, 404, 422, 500) and `$ref` them across all operations
- Add `TenantJwt` security to list events, list attempts, list tenants
- Set `AdminApiKey`-only security on portal and token endpoints
- Remove unsupported query params from list tenants (`order_by`, `created_at[gte]`, `created_at[lte]`)
- Add `azure_servicebus` and `gcp_pubsub` to destination type enum
- Remove `successful_at` from Event schema (not in implementation)

### DIFF
--- a/docs/apis/openapi.yaml
+++ b/docs/apis/openapi.yaml
@@ -2,7 +2,8 @@ openapi: 3.1.0
 info:
   title: Outpost API
   version: "0.0.1"
-  description: The Outpost API is a REST-based JSON API for managing tenants, destinations, and publishing events.
+  description: |
+    The Outpost API is a REST-based JSON API for managing tenants, destinations, and publishing events.
   contact:
     name: Outpost Support
     email: support@hookdeck.com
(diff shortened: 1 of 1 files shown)


leggetter:
This will require SDKs to be regenerated. Does this reflect th deployed API structure? if so, I'll do a patch sdk release.

alexluong:
@leggetter yes for the most part 90% of the changes here are OpenAPI spec/SDK stuff. The API is already working. Only small validation enhancements from the API itself, should be safe.

PR #732, opened 2026-03-06
feat: Refactor API query params to accept array using bracket notation & add array support for more params

alexbouchardd:


internal/apirouter/log_handlers.go
```diff
@@ -293,25 +277,30 @@ func (h *LogHandlers) listAttemptsInternal(c *gin.Context, tenantID string, dest
...
 	if !ok {
 		return
 	}
+	tenantID := ""
+	if len(tenantIDs) == 1 {
+		tenantID = tenantIDs[0]
+	}
 	eventID := c.Param("event_id")
 	event, err := h.logStore.RetrieveEvent(c.Request.Context(), logstore.RetrieveEventRequest{
 		TenantID: tenantID,
```
alexluong: hmm this feels a bit awkward, I wonder if there's a better way to handle `tenant_id` query params, at least for the retrieve event/attempt endpoints

internal/tenantstore/redistenantstore/redistenantstore.go
```diff
@@ -289,6 +290,14 @@ func (s *store) ListTenant(ctx context.Context, req driver.ListTenantRequest) (*
...
 
 	baseFilter := "@entity:{tenant} -@deleted_at:[1 +inf]"
+	if len(req.ID) > 0 {
+		// RediSearch TAG: @id:{val1|val2} matches any of the IDs; escape \ | } in values
+		escaped := make([]string, len(req.ID))
+		for i, id := range req.ID {
+			escaped[i] = strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(id, `\`, `\\`), "|", `\|`), "}", `\}`)
+		}
+		baseFilter += " @id:{" + strings.Join(escaped, "|") + "}"
+	}
```
alexluong, line 305: I'm not 100% sure this is going to be good enough. My concern is that tenant ID is user-provided that may contain special characters (`-` is negation) that may lead to issues. How confident are you about this implementation as a whole?

Should I spend a bit of time to try to break it?

internal/apirouter/pagination.go
```diff
@@ -30,6 +30,15 @@ func CursorToPtr(cursor string) *string {
 	return &cursor
 }
 
+// PrevCursorForResponse returns the prev cursor to include in a pagination response.
+// When no cursor was provided in the request (first page), prev must not be set (returns nil).
+func PrevCursorForResponse(noCursorInRequest bool, prevFromBackend string) *string {
```
alexluong: I'm not 100% sure what problem this solves. In the first page request, the prev cursor would be empty, which `CursorToPtr("") -> nil`. I checked production API and it behaves correctly from what I can tell.

internal/apirouter/log_handlers_test.go
```diff
@@ -971,7 +1071,7 @@ func TestAPI_Attempts(t *testing.T) {
 				{Event: e, Attempt: a},
 			}))
 
-			req := httptest.NewRequest(http.MethodGet, "/api/v1/attempts/a1?include=response_data", nil)
+			req := httptest.NewRequest(http.MethodGet, "/api/v1/attempts/a1?include[0]=response_data", nil)
```
alexluong: small nit but I don't think these changes should be relevant, especially if sending single item, it's probably better to not include `[0]` anyway

alexluong: the `include` query param is a bit tricky, just want to highlight this case, do we expect `?include[0]=...&include[1]=...`?

Maybe? Just feels a bit awkward but that's purely personal.

docs/apis/openapi.yaml
```diff
@@ -2377,12 +2377,30 @@ paths:
         - AdminApiKey: []
```
alexluong, line 2387: OpenAPI spec missing `id[]` param for GET /tenants

internal/apirouter/pagination.go
```diff
@@ -94,10 +103,13 @@ func ParseOrderBy(c *gin.Context, allowedValues []string) (string, *ErrorRespons
...
-// Returns the parsed array in index order. Supports both numeric indices [0], [1] and
-// simple repeated params field[]=a&field[]=b.
-func ParseArrayParam(c *gin.Context, fieldName string) []string {
+// ParseArrayQueryParam parses array query parameters. Supports three formats:
+//   - Indexed bracket notation: field[0]=a&field[1]=b
+//   - Unindexed bracket notation: field[]=a&field[]=b
+//   - Plain repeated params: field=a&field=b (fallback when no bracket notation present)
+//
+// Returns nil if the parameter is absent entirely.
+func ParseArrayQueryParam(c *gin.Context, fieldName string) []string {
```
alexluong, line 103: not sure where to say this so I'll mention it here, should we have some sort of upperbound limit for query params in general? Or specific fields

internal/apirouter/pagination.go
```diff
@@ -94,10 +103,13 @@ func ParseOrderBy(c *gin.Context, allowedValues []string) (string, *ErrorRespons
...
 }
 
-// ParseArrayParam parses bracket notation array parameters (e.g., field[0]=a&field[1]=b).
-// Returns the parsed array in index order. Supports both numeric indices [0], [1] and
-// simple repeated params field[]=a&field[]=b.
-func ParseArrayParam(c *gin.Context, fieldName string) []string {
+// ParseArrayQueryParam parses array query parameters. Supports three formats:
+//   - Indexed bracket notation: field[0]=a&field[1]=b
+//   - Unindexed bracket notation: field[]=a&field[]=b
+//   - Plain repeated params: field=a&field=b (fallback when no bracket notation present)
```
alexluong, line 100: Probably should add some tests to verify all 3 formats are supported?

alexluong: Specifically test for single plain param, imo

internal/apirouter/log_handlers_test.go
alexbouchardd: For consistency then yeah, that's what would be expected. You can also do `include=...&include=...` the brackets are not mandatory

alexbouchardd: Agreed

internal/apirouter/pagination.go
alexbouchardd: Or specific fields?

alexluong: yes, like `tenant_id` arr should only accept X number of items, but `event_id` can support more, etc.

PR #744, opened 2026-03-11
fix(sdks): use JSONPath filter expressions in overlay + simplify Go SDK union types

alexluong:
### 1. JSONPath filter expressions in pagination overlay
Replace brittle positional parameter indices with name-based filter expressions (e.g., `parameters[?@.name == 'dir']` instead of `parameters[2]`). The indices broke when #732 added new query params which shifted all parameter positions.

**Before (broken):** `direction` override applied to `next`, `next_cursor` applied to `time[gt]`, etc.
**After (fixed):** all overrides target the correct params by name, immune to reordering.

### 2. Go SDK: replace oneOf unions with plain arrays
Add a Go-specific Speakeasy overlay (`go-array-params-overlay.yaml`) that converts all `oneOf: [string, array]` query params to plain array types. This eliminates verbose tagged union constructors in the generated Go SDK. Affects 16 params across 8 endpoints.

A new `Outpost API (Go)` source is added to `.speakeasy/workflow.yaml` so the overlay only applies to Go — TS and Python are unaffected.

```go
// Before — tagged union constructors
res, err := s.Events.List(ctx, operations.ListEventsRequest{
    TenantID: operations.CreateListEventsTenantIDArrayOfStr(
        []string{"tenant_123", "tenant_456"},
    ).Pointer(),
    Topic: operations.CreateListEventsTopicArrayOfStr(
        []string{"user.created", "user.updated"},
    ).Pointer(),
})

// After — plain slices
res, err := s.Events.List(ctx, operations.ListEventsRequest{
    TenantID: []string{"tenant_123", "tenant_456"},
    Topic:    []string{"user.created", "user.updated"},
})
```

## Test plan

- [x] `speakeasy overlay validate` passes for both overlays
- [x] `speakeasy overlay apply` confirms oneOf removed, correct param targeting
- [x] `speakeasy run -t outpost-go` — compiles, lints, tests pass
- [x] TS/Python sources unchanged — no DX regression

### DIFF
--- a/.speakeasy/workflow.yaml
+++ b/.speakeasy/workflow.yaml
@@ -10,6 +10,16 @@ sources:
             - location: ./sdks/schemas/error-types.yaml
         registry:
             location: registry.speakeasyapi.dev/hookdeck-dev/outpost/outpost-api
+    Outpost API (Go):
+        inputs:
+            - location: ./docs/apis/openapi.yaml
+        overlays:
+            - location: ./sdks/schemas/speakeasy-modifications-overlay.yaml
+            - location: ./sdks/schemas/security-collapse-overlay.yaml
+            - location: ./sdks/schemas/error-types.yaml
+            - location: ./sdks/schemas/go-array-params-overlay.yaml
+        registry:
+            location: registry.speakeasyapi.dev/hookdeck-dev/outpost/outpost-api
     Outpost API (Python):
         inputs:
             - location: ./docs/apis/openapi.yaml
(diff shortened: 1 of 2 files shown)


.speakeasy/workflow.yaml
```diff
@@ -10,6 +10,16 @@ sources:
             - location: ./sdks/schemas/error-types.yaml
         registry:
             location: registry.speakeasyapi.dev/hookdeck-dev/outpost/outpost-api
```
leggetter, line 12: @alexluong - I'm not a speakeasy expert, but because we've added `Outpost API (Go)`, is this section now redundant?

leggetter: Ah, it's still used by TypeScript.

PR #750, opened 2026-03-13
fix(sdks): extract shared pagination config from Python-only overlay

alexluong:
## Summary

The pagination cursor config (`x-speakeasy-pagination`) was bundled inside the Python-specific overlay, which only applies to the Python SDK source. After #744 split the Speakeasy sources by language, Go and TS SDKs lost auto-pagination support.

- Extract pagination config into a new shared `pagination-config-overlay.yaml` applied to all three sources
- Keep only Python builtin name overrides (`dir`→`direction`, `next`→`next_cursor`, `prev`→`prev_cursor`) in `python-pagination-fixes-overlay.yaml`
- Both overlays pass `speakeasy overlay validate`

## Test plan

- [x] `speakeasy overlay validate` passes for both overlays
- [x] `speakeasy overlay apply` confirms 4 pagination configs applied
- [ ] Re-run SDK generation after merge to verify pagination restored in Go/TS

### DIFF
--- a/.speakeasy/workflow.yaml
+++ b/.speakeasy/workflow.yaml
@@ -8,6 +8,7 @@ sources:
             - location: ./sdks/schemas/speakeasy-modifications-overlay.yaml
             - location: ./sdks/schemas/security-collapse-overlay.yaml
             - location: ./sdks/schemas/error-types.yaml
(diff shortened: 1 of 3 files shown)


leggetter:
Good spot, @alexluong 👍

leggetter (approved):

PR #814, opened 2026-04-09
feat: Add new docs content

alexbouchardd:


leggetter:
@alexbouchardd - should we try to introduce some consistency about how we name the OSS vs managed products?

Doc tags vs. docs directory

- `oss_only` but we put the docs under `self-hosting`.
- `managed_only`, should we put the content under `managed` or `hosted`

How about consistently going with `oss` and `managed`?

alexbouchardd:
@leggetter Fair point, but I think we should go with "Self-hosted" and "Managed" because that's really the distinguishing factor from a docs perspective. Both are technically open source since it's the same code

garethx (approved):

PR #832, opened 2026-04-13
feat(idgen): add configurable prefix for operator event IDs

alexluong:
## Summary

- Add `IDGEN_OPERATOR_EVENT_PREFIX` config to support prefixed IDs for operator events
- Add `OperatorEvent()` method to `IDGenerator` following the same pattern as Event/Destination/Attempt
- Update `opevents.Emitter` to use `idgen.OperatorEvent()` instead of `idgen.String()`

## Test plan

- [x] Unit tests added for `OperatorEvent()` (default UUID, prefix, nanoid+prefix, independence from event prefix)
- [ ] Verify operator events emitted with configured prefix
- [ ] Verify backward compat: no prefix by default


alexbouchardd:
I don't believe that's necessary given that those IDs are internal and we can decide on the ID format. I would close

PR #843, opened 2026-04-17
refactor: deduplicate event rows in ClickHouse

alexluong:
## Summary

- Skip event row insert for retry attempts (`attempt_number > 1`) across all logstore backends
- Deduplicate ListEvent/ListAttempt results client-side: fetch, dedupe by ID, re-fetch if short of limit. Avoids `LIMIT 1 BY` / `GROUP BY` perf penalty on large result sets.
- Fix metrics test dataset to use realistic `attempt_number=1` for unique events

## Test plan

- [x] `TestEventDedup` integration test (fanout, retries, legacy duplicates)
- [x] Conformance suite passes (chlogstore, pglogstore, memlogstore)

### DIFF
--- a/internal/logstore/chlogstore/chlogstore.go
+++ b/internal/logstore/chlogstore/chlogstore.go
@@ -111,6 +111,10 @@ type eventWithPosition struct {
 	eventTime time.Time
 }
 
+func (e eventWithPosition) cursorPosition() string {
+	return fmt.Sprintf("%d::%s", e.eventTime.UnixMilli(), e.Event.ID)
+}
+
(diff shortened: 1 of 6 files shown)


alexbouchardd (commented):
We're missing logic to handle the fan-out scenario, only one of the delivery task should be responsible for the event insert

alexbouchardd:
@alexluong The same dedup logic should apply for attempts, really, we have the same problem there we just haven't seen it because duplicates are rarer

alexluong:
@alexbouchardd yes it's already applied for Attempts as well

alexbouchardd (approved):

Issue #663, opened 2026-01-29
retrymq limit & DLQ


alexluong:
When a retry task's executor fails (e.g., event not found in logstore, transient errors), the message sits in the queue and becomes visible again after a fixed 30s visibility timeout. This repeats indefinitely with no limit.

## Problems

- A permanently failing retry message cycles forever with no cap
- No dead-letter path to detect or surface stuck messages
- Fixed visibility timeout on re-fetch failures — no backoff between attempts

The underlying queue already tracks receive count and supports per-message visibility changes, so the primitives are there.

### Max receive count
What should the default be?

> **Suggestion:** 5 internal re-fetch attempts before giving up. This is separate from the delivery retry max limit, which controls how many times we re-deliver to the destination.

### Backoff on re-fetch
Should we apply exponential backoff on internal failures (e.g., 30s → 60s → 120s), or is a fixed interval fine since these are typically short-lived transient issues?

### What happens when max is exceeded
> **Suggestion:** Route to a DLQ. Gives observability into stuck messages and the ability to replay them.

### Configuration
> **Suggestion:** Expose as `retrymq` config, similar to how `deliverymq` is configured. e.g., `RETRYMQ_MAX_RECEIVE_COUNT`, `RETRYMQ_VISIBILITY_TIMEOUT_SECONDS`.

samadalishah:
Hello @alexluong,

Is there any update on this issue? 

I am running **v0.14.0** using destination type **webhook** only and currently have a scenario where some events failed to publish to log SQS queue with the following error:
```
failed to publish attempt log

pubsub (code=InvalidArgument): operation error SQS: SendMessageBatch, https response error StatusCode: 400, RequestID: SOME_ID, AWS.SimpleQueueService.BatchRequestTooLong: Batch requests cannot be longer than 1048576 bytes. You have sent 1198056 bytes.
```

And ever since the retry keeps on picking the same event and the observability logs are growing with `event not found in logstore, will retry`

My questions are: 
1. Do we always send the events to SQS queues in batches even if the `DELIVERY_MAX_CONCURRENCY` and `LOG_MAX_CONCURRENCY` is set to **1**? 
2. What would be the recommendation in this case to avoid the retries and not to miss sending Webhooks to destination target URLs?

alexluong:
Hey @samadalishah, apologies for the delayed response.

[...]

PR #845, opened 2026-04-20
fix: enforce SQS 1 MB batch byte limit

alexluong:
## Summary

gocloud.dev's SQS topic driver batches `Send()` calls via `SendMessageBatch` (up to 10 messages per batch) but does **not** enforce SQS's 1 MB total batch size limit — it relies on SQS to reject oversized batches at the API level.

When log entries carry large payloads (e.g., large request/response bodies), the combined batch can exceed 1 MB, causing SQS to reject the entire batch with `BatchRequestTooLong`. This failure cascades: the log entry never gets persisted, and the retry scheduler can't find a prior attempt in the logstore, triggering an infinite retry loop (see #663).

This PR sets `MaxBatchByteSize: 1_048_576` on the gocloud `TopicOptions` so the batcher splits oversized batches **before** they reach SQS. If a single message exceeds 1 MB, gocloud returns `ErrMessageTooLarge` instead of a cryptic SQS batch error.

## Changes

- Pass `TopicOptions` with `BatcherOptions.MaxBatchByteSize` set to 1 MB when opening SQS topics in `queue_awssqs.go`

## Test plan

- [x] Verified compilation
- [ ] Deploy to staging with SQS backend and send events with large payloads (>100 KB) at high concurrency to confirm batches are split correctly

### DIFF
--- a/internal/mqs/queue_awssqs.go
+++ b/internal/mqs/queue_awssqs.go
@@ -13,7 +13,13 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"gocloud.dev/pubsub"
 	"gocloud.dev/pubsub/awssnssqs"
+	"gocloud.dev/pubsub/batcher"
 )
+
+// sqsMaxBatchBytes is the maximum total payload size for an SQS SendMessageBatch
+// request. SQS rejects batches exceeding 256 KB per message and 1 MB per batch.
+// Setting this on the gocloud batcher prevents oversized batches from being sent.
+const sqsMaxBatchBytes = 1_048_576 // 1 MB
 
 type AWSSQSConfig struct {
 	Endpoint                  string // optional - dev-focused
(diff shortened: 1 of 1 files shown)


alexbouchardd (approved):

internal/mqs/queue_awssqs.go
```diff
@@ -13,8 +13,14 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"gocloud.dev/pubsub"
 	"gocloud.dev/pubsub/awssnssqs"
+	"gocloud.dev/pubsub/batcher"
 )
 
+// sqsMaxBatchBytes is the maximum total payload size for an SQS SendMessageBatch
+// request. SQS rejects batches exceeding 256 KB per message and 1 MB per batch.
```
samadalishah, line 20: nit: Based [on the docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-messages.html), 
> The minimum message size is 1 byte (1 character). The maximum is 1,048,576 bytes (1 MiB).

1 single message can also be of size 1MB now. So both single message or a batch of 10 messages should be max 1MB

alexluong: Hi @samadalishah, thanks for checking in and reviewing the PR!

I think 1MB is a natural limitation of SQS, so if there's a single message exceeding 1MB then that message will not be able to be sent to SQS. I'm not sure if there's anything we can do in this regards here.

samadalishah: Yes, all good here! 👍
I think this should be considered for issue #663 

As we are adding `Event, Attempt, Destination` in the **LogEntry**, the attempt can sometimes be huge (based on my error example, the webhook destination in case of failure responded with 13K lines of HTML 😅). So either the Attempt's ResponseData should be truncated or the message should be cleaned from `deliverymq-retry` but again related to the issue not this PR.

alexluong: noted, and definitely something we can consider supporting better. You brought up a good point that Attempt's ResponseData is something we need to consider in terms of size limit. cc @alexbouchardd

PR #883, opened 2026-04-29
feat: Add API ref content and openapi schema updates

alexbouchardd:


docs/apis/openapi.yaml
```diff
@@ -1812,7 +2045,7 @@ components:
       properties:
         id:
           type: string
-          description: Optional. A unique identifier for the event. If not provided, a UUID will be generated. A stable id enables publish idempotency (see `duplicate` in the response) and is the value receivers should use to deduplicate webhook deliveries — by default sent as the `X-Outpost-Event-Id` header (prefix configurable).
+          description: Optional. A unique identifier for the event. If not provided, a UUID will be generated.
```
alexluong: Just wanna highlight the UUID part. I wonder if we should be explicit about it given we support different ID template?

docs/apis/openapi.yaml
```diff
@@ -2221,8 +2456,10 @@ components:
       type: object
       properties:
         granularity:
-          type: string
-          description: The granularity used for time bucketing. Absent when none was specified.
+          type:
+            - string
+            - "null"
+          description: The granularity used for time bucketing, or null if none was specified.
```
alexluong: We changed this recently because of SDK issue: https://github.com/hookdeck/outpost/pull/867

alexluong (changes requested):

docs/apis/openapi.yaml
```diff
@@ -1320,7 +1320,7 @@ components:
       properties:
         id:
           type: string
-          description: Optional user-provided ID. A UUID will be generated if empty.
+          description: Optional user-provided ID. A ID will be generated if empty.
```
alexluong: "An ID" is more correct I think

same comment applies to the remaining sentences

alexluong (approved):

PR #894, opened 2026-05-02
feat(logging): add node identifier to log lines

### LINKED ISSUE #307: Add Node Identification to Log Lines
Add a unique identifier for each service instance to log lines to help with debugging in multi-node deployments. Consider using standard fields like hostname or container ID, or environment variables that work across different deployment methods (not k8s specific). This will make it easier to trace issues across logs when running multiple instances of the same service.

`os.Hostname()` provides a reliable node identifier across most modern deployment platforms:
- Kubernetes: returns pod name (e.g. outpost-api-5bbd447f65-x2444)
- Docker: returns container ID
- Cloud VMs: returns instance hostname (usually includes instance ID)
- ECS: returns task/container ID

Cases where hostname alone might not be sufficient:
1. Bare metal/VMs with generic hostnames (web-01, web-02)
2. Local development running multiple instances on same host
3. Platforms that allow hostname customization which might break uniqueness

For our current deployment targets, `os.Hostname()` should be sufficient as a node identifier. We can revisit if we encounter specific use cases where the hostname doesn't provide enough context for debugging.

mvanhorn:
Closes #307.

Adds an `os.Hostname()`-derived `node` field to every log line so multi-node deployments can be traced from log output alone. Per the issue body, `os.Hostname()` returns sensible identifiers across the relevant deployment targets:

- Kubernetes: pod name (e.g. `outpost-api-5bbd447f65-x2444`)
- Docker: container ID
- ECS: task/container ID
- Cloud VMs: instance hostname

## Implementation

The hostname is captured once inside `makeLogger` and attached as a default zap field via `zapLogger.With(zap.String("node", hostname))`. Both the main logger and the audit logger pick this up because they share the same construction path. `otelzap` preserves zap `With` fields, so `Ctx(ctx)` derivations also carry the node identifier with no additional wiring.

If `os.Hostname()` returns an error, the field falls back to `"unknown"`. This matches the silent-fallback pattern in `internal/redislock/redislock.go:136` and avoids failing logger construction on observability metadata.

## Verification

- `go build ./...`: pass
- `go vet ./...`: pass
- `go test ./internal/logging/...`: pass (no existing test files; the change is structural metadata)

## Diff

A single 6-line addition in `internal/logging/logger.go`. No public surface change.

### DIFF
--- a/internal/logging/logger.go
+++ b/internal/logging/logger.go
@@ -74,7 +74,7 @@ func makeLogger(logLevel string) (*otelzap.Logger, error) {
 	if err != nil {
 		hostname = "unknown"
 	}
-	zapLogger = zapLogger.With(zap.String("node", hostname))
+	zapLogger = zapLogger.With(zap.String("host.name", hostname))
 
 	return otelzap.New(zapLogger,
 		otelzap.WithMinLevel(level),
--- a/internal/logging/logger.go
+++ b/internal/logging/logger.go
@@ -2,6 +2,7 @@ package logging
 
 import (
 	"context"
+	"os"
 
 	"github.com/uptrace/opentelemetry-go-extra/otelzap"
 	"go.uber.org/zap"
@@ -69,6 +70,11 @@ func makeLogger(logLevel string) (*otelzap.Logger, error) {
 	if err != nil {
 		return nil, err
 	}
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown"
+	}
+	zapLogger = zapLogger.With(zap.String("node", hostname))
 
 	return otelzap.New(zapLogger,
 		otelzap.WithMinLevel(level),


internal/logging/logger.go
```diff
@@ -69,6 +70,11 @@ func makeLogger(logLevel string) (*otelzap.Logger, error) {
 	if err != nil {
 		return nil, err
 	}
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown"
+	}
+	zapLogger = zapLogger.With(zap.String("node", hostname))
```
alexluong: Let do `host.name` to fit with [OTEL convention](https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/) and also avoid "node" semantic which could be confused witk "k8s node".

```suggestion
	zapLogger = zapLogger.With(zap.String("host.name", hostname))
```

mvanhorn: Done in 2452d49: switched to `host.name` to match the OTEL host semantic convention.

alexluong (approved):
LGTM! Thanks @mvanhorn for the contribution. Welcome, and please let me know if there's anything I can help with Outpost!

Issue #612, opened 2026-01-07
Enable destinations in the Update Destination API


laurenzlong:
**Is your feature request related to a problem? Please describe.**
It would be really nice if updating a destination also enabled it if it was previously disabled. This can either be opt-in behavior via a query param, or automatic behavior. 

**Describe the solution you'd like**
If it was a query param, then something like `enable=true` could work. 

**Describe alternatives you've considered**
The current solution is to call the enable destination right after the update destination request, which is a viable workaround. 

**Additional context**
N/A

alexbouchardd:
Good feedback but low priority given there's a decent work-arround!

PR #895, opened 2026-05-05
feat(api): enable/disable destinations via Create and Update

### LINKED ISSUE #488: Allow enabling/disabling a destination at creation
Currently, it is possible to enable or disable a destination, but only after it has been created. This means you need to make two API calls if you want to create a destination that is disabled.

It would be useful to choose the state (enabled/disabled) directly when creating the destination, with "enabled" as the default value.

N/A

**Additional context**
N/A

alexluong:
## Summary

This PR bundles two related but distinct scopes.

### 1. Enable/disable via Create and Update (closes #488, #612)

- **Create** accepts optional `disabled_at` so a destination can be created in a disabled state (e.g. importing a disabled Svix endpoint without a window where traffic could leak to it).
- **Update** accepts optional `disabled_at` (omit / `null` / timestamp) so callers can enable/disable as part of a normal PATCH. The dedicated `/enable` and `/disable` endpoints stay — this is the import-friendly path.
- Update emits an additional `destination disabled` / `destination enabled` audit log alongside `destination updated` when disability state changes.

### 2. Importable timestamps on Create

- **Create** accepts optional `created_at` and `updated_at` so importers can preserve the original timestamps from the source system.
- No validation on these fields — values are accepted as-is. Future timestamps and out-of-order timestamps are allowed; importers from other systems may carry inconsistent data and we'd rather accept it than block a migration.

## Test plan

- [x] `go test ./internal/apirouter/...` — all pass
- [ ] Manual: POST a destination with past `created_at` / `disabled_at`, verify timestamps preserved in response and storage
- [ ] Manual: PATCH \`disabled_at: null\` on a disabled destination, confirm it enables and emits audit log

### DIFF
--- a/docs/apis/openapi.yaml
+++ b/docs/apis/openapi.yaml
@@ -1355,8 +1355,8 @@ components:
           nullable: true
           description: >-
             Optional override for the creation timestamp. Intended for importing
-            destinations from another system. Defaults to the current time when
-            omitted.
+            destinations from another system. Must not be in the future.
+            Defaults to the current time when omitted.
           example: "2024-02-15T10:00:00Z"
         updated_at:
           type: string
@@ -1364,7 +1364,8 @@ components:
           nullable: true
           description: >-
             Optional override for the last-updated timestamp. Intended for
-            importing destinations. Defaults to created_at when omitted.
+            importing destinations. Must not be in the future. Defaults to
+            created_at when omitted.
           example: "2024-02-15T10:00:00Z"
         disabled_at:
           type: string
@@ -1415,8 +1416,8 @@ components:
           nullable: true
           description: >-
             Optional override for the creation timestamp. Intended for importing
-            destinations from another system. Defaults to the current time when
-            omitted.
+            destinations from another system. Must not be in the future.
+            Defaults to the current time when omitted.
           example: "2024-02-15T10:00:00Z"
         updated_at:
           type: string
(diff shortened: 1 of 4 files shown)


alexbouchardd:
Is the expected usage for importing a disabled destination to set `disabled_at: current timestamp`? I wonder if it might not be confusing in the case where disabled_at is in the future. Will it only take effect in the future or as long as a date is set it's disabled?

alexluong:
> I wonder if it might not be confusing in the case where disabled_at is in the future. Will it only take effect in the future or as long as a date is set it's disabled?

No, the current logic is if disabled_at exists -> take effect right away. That said, I reckon it's quite trivial to check the value of that timestamp compared to 'now'.

In what scenario would someone want to set a future timestamp for disabled_at?

alexbouchardd:
> In what scenario would someone want to set a future timestamp for disabled_at?

I don't see the use case, but I'm pointing it out as a potential point of confusion if there's no validation

alexluong:
We can add validation that disabled_at cannot be in the future?

alexbouchardd:
Yeah although that does bring up the question as to why created_at/updated_at could be in the future 🤔

alexbouchardd:
Aside we should probably limit created_at/updated_at input to API key auth?

alexluong:
> Yeah although that does bring up the question as to why created_at/updated_at could be in the future 🤔

fair 😅 

> Aside we should probably limit created_at/updated_at input to API key auth?

hmm, let me think, I don't believe we have a lot of logic that's limited to API key auth vs JWT, but I see the point.

alexluong:
@alexbouchardd updated, ready for another round of review now!

alexbouchardd (approved):

PR #899, opened 2026-05-12
feat(webhook): proxy-aware error handling for forward proxies

alexluong:
## Summary

Webhook deliveries through a forward proxy (`DESTINATIONS_WEBHOOK_PROXY_URL`) now distinguish proxy *infrastructure* failures from destination failures, and translate proxy-reported destination failures into the right destination-attributed error code. No new config; Envoy-specific signals are auto-detected from response headers when present. Implementation is scoped to `destwebhook/`; `destregistry/` exposes only a generic `WrapTransport` hook.

Full feature doc: [`docs/content/features/webhook-proxy.mdoc`](./docs/content/features/webhook-proxy.mdoc).

## Why

Two equally-load-bearing problems today:

- **Misattributed retry budget.** A proxy outage records as a destination failure, burning the destination's retry budget and DLQ slot for a problem it can't fix.
- **Wrong error surface to customers.** Upstream errors that the proxy *reports* (DNS, connection refused, timeout) reach the delivery record as proxy-flavored generic errors instead of the underlying destination problem, and `server: envoy` / `x-envoy-*` headers leak into the response stored against the destination.

## Behavior

- **Proxy auth failure or dial-to-proxy failure** → nack. The queue redelivers; no customer-visible attempt is recorded.
- **Any other proxy-reported failure** (CONNECT 5xx, Envoy-synthesized 5xx on plain-HTTP) → destination attempt with a refined code (`UF` → `connection_refused`, `DC` → `dns_error`, `UT` → `timeout`, etc.; unrecognized → `network_error`).
- **Plain-HTTP success path** — `x-envoy-*` and `server: envoy` headers are stripped before storage. Envoy-synthesized 5xx bodies are dropped.
- **HTTPS path** — byte-transparent. TLS is end-to-end after CONNECT, so the wrapper never inspects or modifies the response. Destinations behind their own Envoy keep their real status, body, and headers.

## Nuances

- **Spoof-resistance on the plain-HTTP path** requires the operator's Envoy to emit `x-envoy-response-flags` with `OVERWRITE_IF_EXISTS_OR_ADD` (must live on `RouteConfiguration`, Envoy rejects it on HCM). Without this, a destination can fake the flag header and force its real body to be dropped.
- **Plain-HTTP destinations behind their own Envoy** lose some observability headers (we strip them). Attribution stays correct because the forward Envoy overwrites the flag. HTTPS destinations are unaffected.
- **Generic (non-Envoy) proxies** get the nack-on-auth-failure / nack-on-dial-failure paths. The Envoy header logic is a no-op when the headers aren't present.

## Pub/Sub redelivery runway

Outpost's default Pub/Sub provisioning gives roughly **5 minutes of redelivery before DLQ** (see the docs page for the exact knobs). A proxy outage under that window is transparent to destinations; longer ones land in the DLQ and need manual replay.

## Test plan

- [ ] Wrapper unit tests against fake proxy: CONNECT 407 → nack; CONNECT 5xx → destination record; synthesized response with each Envoy flag → correct mapped code; real upstream pass-through; HTTPS response with envoy headers → pass-through untouched
- [ ] `ExecuteHTTPRequest` returns `Delivery: nil` only on the infra-error sentinel
- [ ] Pin test on Go's `proxyconnect` error wording (guards stdlib regression)
- [ ] `MapEnvoyResponseFlag` table test
- [ ] Manual smoke test through real Envoy in QA: HTTPS dest, HTTP dest, proxy auth failure, proxy unreachable, destination DNS failure (verify recorded as destination), destination 500 (verify pass-through), destination behind Envoy returning 503 (verify customer sees the 503, not a synthesized error)


alexbouchardd:
Is `DESTINATIONS_WEBHOOK_PROXY_TYPE=envoy` necessary? Can't we infer that envoy is used by the response headers in case of an error?

alexluong:
> Is DESTINATIONS_WEBHOOK_PROXY_TYPE=envoy necessary? Can't we infer that envoy is used by the response headers in case of an error?

hmm good idea, yeah we can do that I think

alexbouchardd (approved):
I would get @mkherlakian review here as well since he knows envoy's quirks well

internal/destregistry/providers/destwebhook/proxytransport_internal_test.go
mkherlakian, line 1: Is it common practice in Go to put tests in the same directory as the code it's testing?

mkherlakian (approved):

internal/destregistry/providers/destwebhook/proxytransport_internal_test.go
alexluong: @mkherlakian yes, especially for unit test it's sort of the only option. Putting it in a different directory would be an anti-pattern.

Also keep in mind Golang enforce an implicit package boundary by directory, so there's that.

PR #910, opened 2026-05-22
feat(publishmq): NATS JetStream source with per-account multi-tenancy

michaeldoehler:
Adds NATS JetStream as a publish-mq source for Outpost. Outpost reads
events from one or more pre-provisioned JetStream consumers; streams
and consumers are operator-owned. The driver is built for the common
SaaS pattern of *one NATS Account per Outpost tenant*.

## Why

NATS JetStream is a popular event bus in self-hosted Kubernetes / on-prem
setups, and a natural source for Outpost's webhook-fanout role. The driver
slots into the existing PublishMQ interface alongside RabbitMQ, AWS SQS,
GCP Pub/Sub and Azure Service Bus.

Scope is intentionally limited to publish-mq only — Outpost does not run
its internal delivery/log queues on NATS. That keeps the surface area
small and avoids dragging JetStream into Outpost's auto-provisioning
contract.

## Design highlights

- **Multi-account**: each ` + "`PublishNATSAccountConfig`" + ` holds its own credentials
  file and is dialled on its own NATS connection. Pull loops run in
  parallel and feed into a single ` + "`Subscription.Receive`" + ` channel.
- **Tenant isolation**: when an account sets ` + "`tenant_id`" + ` (recommended),
  Outpost overrides the payload's ` + "`tenant_id`" + ` so a publisher with creds
  for Account A can only produce events for the mapped tenant.
- **Dynamic accounts**: an optional ` + "`accounts_dir`" + ` is watched via
  ` + "`fsnotify`" + ` with a 250ms debounce. New tenant subdirectories trigger a
  new connection without restarting Outpost; removed dirs drain the
  connection. Provisioning flow: mint NATS account JWT → push to resolver
  → drop ` + "`meta.yaml`" + ` + ` + "`.creds`" + ` into ` + "`accounts_dir`" + ` → Outpost picks it up.
- **Auth**: ` + "`credentials_file`" + ` (.creds) is the primary mode. It is
  optional — empty means no-auth (valid for trusted-network setups or
  token-via-URL).
- **Operator-owned infra**: streams and durable consumers must exist
  before Outpost starts. ` + "`Init`" + ` verifies both and fails loudly on any
  missing piece. ` + "`AckWait`" + ` / ` + "`MaxDeliver`" + ` are configured on the consumer.
- **No publish from Outpost**: ` + "`Publish()`" + ` intentionally returns an error;
  JetStream is read-only from Outpost's side.

## Configuration

` + "```yaml" + `
publishmq:
  nats:
    servers:
      - nats://nats:4222
    accounts_dir: /etc/outpost/nats-accounts
    accounts:
      - name: acme
        credentials_file: /etc/outpost/acme.creds
        stream: events
        consumer: outpost
        tenant_id: acme
` + "```" + `

Env vars: ` + "`PUBLISH_NATS_SERVERS`" + ` (comma-separated cluster URLs),
` + "`PUBLISH_NATS_ACCOUNTS_DIR`" + `.

Accounts-dir layout:

` + "```" + `
/etc/outpost/nats-accounts/
├── acme/
│   ├── user.creds      # optional NATS .creds (JWT + NKey seed)
│   └── meta.yaml       # name/stream/consumer/tenant_id
└── globex/
    ├── user.creds
    └── meta.yaml
` + "```" + `

Full guide added at ` + "`docs/content/publishing/publish-from-nats.mdoc`" + `.

## Commit walk-through

- ` + "`feat(mqs): add NATS JetStream queue driver`" + ` — driver skeleton,
  ` + "`NATSConfig`" + `/` + "`NATSAccountConfig`" + `, pull-consumer subscription, tenant
  override, wired into ` + "`mqs.QueueConfig`" + ` and ` + "`NewQueue`" + `.
- ` + "`feat(mqs): NATS accounts directory watcher`" + ` — refactors connection
  storage to a name-keyed map and adds the ` + "`fsnotify`" + `-based watcher
  with dynamic add/remove.
- ` + "`feat(config): wire PublishNATSConfig into publishmq`" + ` — config
  surface area plus ` + "`GetInfraType`" + ` / ` + "`GetQueueConfig`" + ` branches.
- ` + "`feat(cmd/publish): NATS JetStream dev publish helper`" + ` — adds
  ` + "`method=nats`" + ` to the existing publish dev service.
- ` + "`test(mqs): NATS JetStream integration tests + testinfra`" + ` — four
  integration tests (basic, tenant-override, multi-account, watcher),
  ` + "`nats:2.10-alpine`" + ` testcontainer, declare/teardown helpers.
- ` + "`docs: publish-from-nats guide + env/yaml examples + compose`" + ` —
  user-facing guide, ` + "`.env.example`" + `, ` + "`.outpost.yaml.example`" + `,
  ` + "`contributing/mq.md`" + ` update, and a docker-compose example with
  ` + "`nats:2.10-alpine`" + ` + helper script.

## Verification

- ` + "`go build ./...`" + ` + ` + "`go vet ./...`" + ` clean.
- ` + "`go test -run TestIntegrationMQ_NATS -count=1 ./internal/mqs/...`" + `
  passes locally (testcontainer ` + "`nats:2.10-alpine -js`" + `):
  - ` + "`TestIntegrationMQ_NATS`" + ` — basic publish/receive/ack
  - ` + "`TestIntegrationMQ_NATS_TenantOverride`" + ` — payload tenant_id is
    overridden by account config
  - ` + "`TestIntegrationMQ_NATS_MultiAccount`" + ` — two accounts consumed in
    parallel, each tagged with its own tenant
  - ` + "`TestIntegrationMQ_NATS_AccountsDir`" + ` — watcher picks up an
    account dir created after Init

## Dependencies

Promotes ` + "`github.com/nats-io/nats.go`" + ` from indirect to direct (already
present in ` + "`go.sum`" + ` via transitive deps; no version bump). Promotes
` + "`github.com/fsnotify/fsnotify`" + ` from indirect to direct.

## Out of scope

- Internal MQ support (delivery/log). Could be added later but is a
  separate decision around JetStream's storage/replication model.
- Kafka publish-mq (separately tracked in ` + "`contributing/mq.md`" + `).
- Credentials rotation without re-creating the account directory.

Happy to split commits differently or rework any part — the staging is
designed to keep each commit independently reviewable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -97,6 +97,13 @@
 # PUBLISH_AZURE_SERVICEBUS_TOPIC="outpost-publish"
 # PUBLISH_AZURE_SERVICEBUS_SUBSCRIPTION="outpost-publish-sub"
 
+## NATS JetStream
+# Comma-separated NATS server URLs (cluster-aware).
+# PUBLISH_NATS_SERVERS="nats://nats:4222"
+# Per-tenant accounts live in subdirectories of this dir. Watched at runtime.
+# Each subdirectory contains a meta.yaml plus an optional user.creds file.
+# PUBLISH_NATS_ACCOUNTS_DIR="/etc/outpost/nats-accounts"
+
 # ============================== Others ==============================
 
 # Portal
--- a/.env.test
+++ b/.env.test
@@ -4,6 +4,7 @@ TEST_CLICKHOUSE_URL="localhost:39000"
 # MQs
 TEST_RABBITMQ_URL="localhost:35672"
 TEST_KAFKA_URL="localhost:39092"
+TEST_NATS_URL=""
 TEST_LOCALSTACK_URL="localhost:34566"
 TEST_GCP_URL="localhost:38085"
 TEST_AZURE_SB_CONNSTRING="Endpoint=sb://localhost;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;"
--- a/.outpost.yaml.example
+++ b/.outpost.yaml.example
@@ -42,6 +42,28 @@ mqs:
     delivery_queue: "outpost-delivery" # SQS queue name for delivery events
     log_queue: "outpost-log" # SQS queue name for log events
 
+# Publish Message Queue (optional)
+# Configure one provider to ingest events from an external source instead
+# of (or in addition to) the Publish API. Only one provider may be set.
+# publishmq:
+#   rabbitmq:
+#     server_url: "amqp://guest:guest@localhost:5673"
+#     exchange: "outpost"
+#     queue: "publish"
+#   nats:
+#     servers:
+#       - "nats://nats:4222"
+#     # Watched directory of per-tenant account subdirectories.
+#     # Each subdirectory holds a meta.yaml and optional user.creds.
+#     accounts_dir: "/etc/outpost/nats-accounts"
+#     # Static list of accounts (combined with accounts_dir if both are set).
+#     accounts:
+#       - name: "acme"
+#         credentials_file: "/etc/outpost/acme.creds"
+#         stream: "events"
+#         consumer: "outpost"
+#         tenant_id: "acme"  # Outpost stamps this on every event; overrides payload.
+
 # Application Configuration
 aes_encryption_secret: "" # Secret for AES encryption
 topics: # List of topics to subscribe to
--- a/cmd/publish/declare_handler.go
+++ b/cmd/publish/declare_handler.go
@@ -16,6 +16,8 @@ func handleDeclare(w http.ResponseWriter, r *http.Request) {
 		err = declareGCP()
 	case "rabbitmq":
 		err = declareRabbitMQ()
+	case "nats":
+		err = declareNATS()
 	case "http":
 		log.Println("[*] Declare HTTP - nothing to declare")
 	default:
--- a/cmd/publish/publish_handler.go
+++ b/cmd/publish/publish_handler.go
@@ -23,6 +23,8 @@ func handlePublish(w http.ResponseWriter, r *http.Request) {
 		err = publishGCP(body)
 	case "rabbitmq":
 		err = publishRabbitMQ(body)
+	case "nats":
+		err = publishNATS(body)
 	case "http":
 		fallthrough
 	default:
--- a/cmd/publish/publish_nats.go
+++ b/cmd/publish/publish_nats.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"log"
+	"os"
+	"time"
+
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nats.go/jetstream"
+)
+
+// Defaults match examples/docker-compose/compose-publish-nats.yml.
+const (
+	natsDefaultURL      = "nats://localhost:4222"
+	natsDefaultStream   = "outpost-publish"
+	natsDefaultSubject  = "outpost.publish"
+	natsDefaultConsumer = "outpost"
+)
+
+func natsURL() string {
+	if v := os.Getenv("PUBLISH_NATS_URL"); v != "" {
+		return v
+	}
+	return natsDefaultURL
+}
+
+func natsSubject() string {
+	if v := os.Getenv("PUBLISH_NATS_SUBJECT"); v != "" {
+		return v
+	}
+	return natsDefaultSubject
+}
+
+func natsStream() string {
+	if v := os.Getenv("PUBLISH_NATS_STREAM"); v != "" {
+		return v
+	}
+	return natsDefaultStream
+}
+
+func natsConsumer() string {
+	if v := os.Getenv("PUBLISH_NATS_CONSUMER"); v != "" {
+		return v
+	}
+	return natsDefaultConsumer
+}
+
+func natsCredsFile() string {
+	return os.Getenv("PUBLISH_NATS_CREDS")
+}
+
+func publishNATS(body map[string]interface{}) error {
+	log.Printf("[x] Publishing NATS JetStream")
+
+	opts := []nats.Option{nats.Name("outpost-publish-dev")}
+	if creds := natsCredsFile(); creds != "" {
+		opts = append(opts, nats.UserCredentials(creds))
+	}
+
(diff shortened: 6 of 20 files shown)


michaeldoehler:
Thanks for the review. Pushed `59fd97b8` addressing six of the seven points:

- **`queue_nats.go:438` busy-spin** — added a 250ms backoff between non-fatal `iter.Next()` errors. Loop stays responsive to shutdown via `select` on `s.done`.
- **`nats_accounts.go:49` silent stat errors** — only `os.ErrNotExist` is treated as "skip"; other errors propagate so operators see permission/IO issues.
- **`nats_accounts.go:155` timer leak** — switched to a single reusable timer with `Reset`/`Stop` + proper channel drain.
- **`testutil/nats.go:70` substring error match** — replaced with `errors.Is(err, jetstream.ErrStreamNotFound)`.
- **`testinfra.go:72` scheme detection** — now `strings.HasPrefix` checking `nats://` and `tls://`.
- **`queue_nats.go:278` dropped reconcile errors** — `addAccount` failures during directory reconciliation are now logged with the account name so bad creds / missing streams don't silently disappear.

On the remaining point:

- **`testinfra/nats.go:89` unsynchronized `cfg.cleanupFns` append** — the race detector actually flagged a sibling issue: a classic double-checked-locking race in `EnsureNATS` (read of `cfg.NATSURL` outside `sync.Once.Do`). That's now fixed by moving the check inside `Once.Do`, and tests pass under `-race`. The broader concern about concurrent `append(cfg.cleanupFns, …)` across different `Ensure*` helpers (RabbitMQ + Kafka + NATS in parallel) is a real pre-existing pattern in this file — same shape in `rabbitmq.go`, `kafka.go`, etc. Happy to follow up with a separate PR that mutex-protects `Config.cleanupFns` (or replaces it with a registration method) so all the test-infra helpers benefit consistently — felt out of scope for this one.

alexluong:
Hi @michaeldoehler, thanks for opening the PR, this was quite interesting. My main concern is the complexity around multi-tenancy via multi-account NATS setup. Can you elaborate more on this design decision? Is this a current use case within your system?

From Outpost POV, we see the publisher to Outpost as 1 entity (your service), which would then result in 1 publishmq only. I'd love to understand the use case where per-account publish queues are necessary for each tenant?

michaeldoehler:
Hi @alexluong, that's a fair question and your framing is right for the common case. Let me sketch what we're working with, because I think once the picture's on the table the rest makes sense.

We're building a B2B platform where each customer has their own workspace, end to end. Data in one place, identity in another, billing in a third, kept apart at the boundary in a way the systems themselves can enforce, not by convention. When a customer onboards, that boundary extends down to messaging too: each customer gets their own NATS account with their own credentials.

So from Outpost's perspective it really is "one publisher, one publishmq" per customer. The wrinkle is that there are many customers and the natural thing for Outpost to do is consume across all of them. That's where the per-account connections and the override come from. They aren't adding complexity for its own sake, they're how we let Outpost speak to the boundary that already exists.

The reason we lean on NATS accounts rather than a shared account with `tenant_id` in the payload is who's doing the publishing. Some of our own services produce events, but tenants can also install third-party apps from a marketplace, and some run their own integrations into our bus. We can't honestly say all of those publishers live inside the same trust boundary. If we trusted payload `tenant_id` we'd be promising the security property "no publisher will ever spoof a tenant" and we'd have no way to enforce it. With per-account credentials, NATS enforces it on connect, before a single byte of payload is read. The override in the driver follows from the same idea: the auth boundary is the source of truth, the payload is just data.

We also use account-level resource limits (bytes/sec, max connections, JetStream quota) to back our plan tiers. Subject conventions plus external counters can fake some of that, but it leaks too easily to be reliable.

Reading the PR top to bottom, the multi-conn and watcher pieces all come back to that single shape: the tenant boundary is already a NATS account, Outpost needs to consume across all of them, and our provisioning flow doesn't wait for an Outpost restart, it drops a creds file plus meta.yaml into a directory and expects everything downstream to keep up.

I tried to keep that from leaking into the simple case. With one account in config you get the RabbitMQ shape exactly: one connection, no watcher, no override, no fan-in. Anything new sits behind the NATS provider branch, the other publish-mq paths and the infra-mq adapters aren't touched.

The maintenance concern is fair though. Multi-conn lifecycle, the watcher, the override logic, that's genuine surface area to carry. If a smaller first step would land more comfortably, I can break this into two PRs: a single-account driver matching the RabbitMQ shape first, then the multi-account and directory bits as a follow-up that gets evaluated on its own. The commits already roughly map to that split.

If neither of those sounds right, point me at what feels heaviest from your side and I'll see what can be simplified without losing the property we need.

michaeldoehler:
Any Updates? We using this already in production, would be nice if we need to go with our fork or if this goes into a future version.

alexluong:
Hi @michaeldoehler, sorry for the delay. The single-account driver is something we'd be comfortable accepting — happy to review that as its own PR, and the split you suggested along the existing commits sounds right.

As for the multi-account and directory-watching aspect, I don't think it's a good fit for Outpost. Your security requirement makes sense — tenant identity should come from the auth boundary, not the payload. But the design principle we want to hold is that Outpost consumes from **one** publish source over one connection, and the fan-in across tenant accounts happens upstream of us. Per-account connections would grow with tenants × Outpost replicas, and the accounts directory puts account lifecycle and credential distribution inside Outpost — that's control-plane work your provisioning system already owns and would do better.

Two ways to get that fan-in on your side, keeping your account-per-tenant boundary exactly as it is:

1. **NATS-native**: a central account runs one aggregate stream that *sources* from each tenant's stream, with a per-source subject transform (NATS 2.10+) prefixing the tenant name — `events.order_paid` arrives as `tenant.acme.events.order_paid`. The transform is operator config, so the tenant token is just as unforgeable as your per-account credentials. Onboarding a tenant is one stream-update call adding a source.
2. **A thin forwarder of your own**: a small service that consumes each tenant account (it knows which credentials it dialed with), stamps `tenant_id`, and republishes into a single trusted stream. Same override logic your PR has, but living next to your provisioning flow instead of inside Outpost.

Either way, Outpost consumes one stream with the single-account driver and the property you need is preserved end to end. If you go the subject-transform route, we could potentially support a small driver option — something like `tenant_id_from_subject` — so Outpost stamps `tenant_id` from the subject token directly and you don't need a forwarder at all. Happy to discuss that as a follow-up once the base driver lands.

Would either of these fit your provisioning flow?

PR #912, opened 2026-05-25
feat(api): support wildcard topic subscriptions

### LINKED ISSUE #908: Support wildcard patterns in destination topic subscriptions
## Summary

Today destinations subscribe to topics by exact match (or `"*"` for all). Customers want to subscribe to topic families without enumerating every topic — e.g. `user.*`, `*.created`, `order.*.completed`.

## Current behavior

Topic matching is exact equality. The only pattern supported is the single-string `"*"` meaning subscribe-all.

## Proposed behavior

Treat `*` inside a topic string as a wildcard matching any run of characters (separator-agnostic — Outpost doesn't enforce a topic separator).

Examples:
- `user.*` matches `user.created`, `user.profile.updated`
- `*.created` matches `user.created`, `order.created`
- `order.*.completed` matches `order.payment.completed`

Exact-match strings (no `*`) keep current behavior.

Abdulmumin1:
## Summary

adds support for wildcard patterns in destination topic subscriptions.

destinations can now subscribe to topic families using `*` inside a topic string, for example:

- `user.*`
- `*.created`
- `order.*.completed`

exact topic subscriptions keep the existing behavior, and `"*"` still means subscribe to all topics.

closes #908

## Behavior

`*` inside a topic string matches any run of characters, separator-agnostic.

examples:

- `user.*` matches `user.created`
- `user.*` matches `user.profile.updated`
- `*.created` matches `user.created` and `order.created`
- `order.*.completed` matches `order.payment.completed`

published event topics are still exact topic strings. this only changes destination subscription matching/validation.

## Validation

if available topics are configured, wildcard patterns must match at least one available topic.

for example, with available topics:

`user.created`
`user.deleted`

this is valid:

`user.*`

this is rejected:

`order.*`

if no available topics are configured, wildcard patterns are accepted, same as other destination topics today.

## Implementation notes

topic matching is on the delivery hot path, so this avoids regex and avoids compiling/caching patterns.

the matcher uses a small allocation-free wildcard algorithm:

- exact strings still short-circuit
- `*` matches any run of characters
- matching is separator-agnostic
- benchmark coverage was added for exact match/miss and pattern match/miss

## Tests

verified with local test infra running:

`TESTINFRA=1 go test ./internal/publishmq -run 'TestEventHandler|TestIntegrationPublishMQEventHandler_Concurrency' -count=1`

`TESTINFRA=1 go test ./internal/models ./internal/tenantstore/memtenantstore ./internal/tenantstore/redistenantstore ./internal/apirouter ./internal/publishmq -count=1`

bench:

`TESTINFRA=1 go test ./internal/models -bench 'BenchmarkTopics_MatchTopic' -benchmem -run '^$'`

result on my machine:

| Benchmark                                      | Time        | Bytes/op | Allocs/op |
|-----------------------------------------------|-------------|----------|-----------|
| BenchmarkTopicsMatchTopic/exactmatch-8        | 34.26 ns/op | 0 B/op   | 0 allocs/op |
| BenchmarkTopicsMatchTopic/exactmiss-8         | 30.93 ns/op | 0 B/op   | 0 allocs/op |
| BenchmarkTopicsMatchTopic/patternmatch-8      | 77.37 ns/op | 0 B/op   | 0 allocs/op |
| BenchmarkTopicsMatchTopic/patternmiss-8       | 93.08 ns/op | 0 B/op   | 0 allocs/op |

### DIFF
--- a/internal/portal/src/common/TopicPicker/TopicPicker.tsx
+++ b/internal/portal/src/common/TopicPicker/TopicPicker.tsx
@@ -1,4 +1,4 @@
-import { useState, useMemo } from "react";
+import { useState, useMemo, useEffect } from "react";
 
 import "./TopicPicker.scss";
 import { Checkbox } from "../Checkbox/Checkbox";
@@ -34,27 +34,50 @@ const findFirstSeparator = (topic: string): string | null => {
   return firstSep;
 };
 
-const topics: Topic[] = (() => {
-  const topicsList = CONFIGS.TOPICS.split(",");
-
-  return topicsList.map((topic) => {
-    const separator = findFirstSeparator(topic);
-
-    return {
-      id: topic,
-      category: separator ? topic.split(separator)[0] : topic,
-    };
-  });
-})();
-
 const TopicPicker = ({
   maxHeight,
   selectedTopics,
   onTopicsChange,
 }: TopicPickerProps) => {
+  // Keep track of any custom topics seen during this component's lifecycle
+  // so they don't disappear if they are temporarily unselected or "Select All" is clicked.
+  const [seenTopics, setSeenTopics] = useState<string[]>([]);
+
+  useEffect(() => {
+    setSeenTopics((prev) => {
+      const next = new Set(prev);
+      let changed = false;
+      for (const t of selectedTopics) {
+        if (t !== "*" && !next.has(t)) {
+          next.add(t);
+          changed = true;
+        }
+      }
+      return changed ? Array.from(next) : prev;
+    });
+  }, [selectedTopics]);
+
+  // Combine statically configured topics with any custom topics already selected
+  const allTopics: Topic[] = useMemo(() => {
+    const configuredTopics = CONFIGS.TOPICS.split(",").filter(Boolean);
+    const combinedSet = new Set([
+      ...configuredTopics,
+      ...seenTopics,
+      ...selectedTopics.filter((t) => t !== "*"),
+    ]);
+
+    return Array.from(combinedSet).map((topic) => {
+      const separator = findFirstSeparator(topic);
+      return {
+        id: topic,
+        category: separator ? topic.split(separator)[0] : topic,
+      };
+    });
+  }, [selectedTopics, seenTopics]);
+
   const [searchQuery, setSearchQuery] = useState("");
   const [expandedCategories, setExpandedCategories] = useState<string[]>(
-    Array.from(new Set(topics.map((topic) => topic.category))),
+    Array.from(new Set(allTopics.map((topic) => topic.category))),
   );
 
   const isEverythingSelected = selectedTopics.includes("*");
@@ -69,7 +92,7 @@ const TopicPicker = ({
 
   // Group topics by category
   const categorizedTopics = useMemo(() => {
-    const filtered = topics.filter((topic) =>
+    const filtered = allTopics.filter((topic) =>
       topic.id.toLowerCase().includes(searchQuery.toLowerCase()),
     );
 
@@ -84,7 +107,7 @@ const TopicPicker = ({
       },
       {} as Record<string, Topic[]>,
     );
-  }, [topics, searchQuery]);
+  }, [allTopics, searchQuery]);
 
   const toggleCategory = (category: string) => {
     setExpandedCategories((prev) =>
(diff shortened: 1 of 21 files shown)


alexluong (commented):
Heya, overall looks great and we should be quite close!

A few notes

1: I see we're not changing the portal (internal/portal) yet. Do you intend for that to be out of scope of this PR?
2: There's this case on the List Destination API where we can run `GET /tenants/X/destinations?topics=user.created`. I wonder what the behavior for this should be with wildcard support.
3: We should add tests to make sure destination with wildcard topics will be matched properly in event delivery (via /publish). I can also help commit the test for this if you prefer.

Abdulmumin1:
awesome @alexluong  about the portal stuff.. i think i might let the team decide on how the ui would evolve. as @alexbouchardd pointed out. the current topic input can't really accommodate it cleanly.

image


i might play around with something.. but i'm not going to be making any decisions on that

alexbouchardd:
You can maybe map the top-level `User` checkbox to `user.*`. In which case, manually selecting all "children" wouldn't select the parent. Selecting the parent could then use a wildcard? Maybe that's too subtle though

A bigger issue with just ignoring it is that the UI may break if the user uses the API to set their topics with a wildcard and then later loads the UI

Abdulmumin1:
hey everyone, ended up playing around with the portal stuff after all to get the ui support the wildcard logic. 

i went with what @alexbouchardd suggested:
- the top-level parent checkbox (like User) now explicitly acts as the toggle for the wildcard (user.*)
- if you manually select user.created, the parent just shows the indeterminate - state. clicking the parent upgrades you to the wildcard and clears out the individual children. 
- wildcard labels actually show the pattern now, e.g., User (user.*).
- you can just search order.* (or any valid glob) and it lets you add it on the fly. 

here’s a quick demo showing the flow:

https://github.com/user-attachments/assets/a7f659c5-4614-4b6b-b898-ecda936504bb


open to corrections on this! ty

alexbouchardd:
Thanks! Sorry for the delay, we have some competing priorities will get back to this soon 👍

Abdulmumin1:
awesome.. available whenever ure ready! ty

internal/portal/src/common/TopicPicker/TopicPicker.tsx
```diff
@@ -94,79 +117,102 @@ const TopicPicker = ({
...
-    } else {
-      const newSelected = new Set([...selectedTopics, ...categoryTopicIds]);
-      onTopicsChange(Array.from(newSelected));
-    }
-  };
-
   const toggleTopic = (topicId: string) => {
     if (isEverythingSelected) {
       selectedTopics = [];
     }
```
alexluong: Not the biggest fan of mutation like this (reassignment of `selectedTopics` which is a prop). Don't think it's a bug but there should be a better way to handle this case?

alexluong (approved):
Heya, appreciate the contribution and your patience. I've got around to reviewing this and everything looks great. Only one small nit on the FE above.

@Abdulmumin1 feel free to address the comment above however you see fit and merge the PR accordingly.

cc @alexbouchardd if you'd like to do a final review on the code / demo

internal/portal/src/common/TopicPicker/TopicPicker.tsx
Abdulmumin1: ahh.. let me fix that.

we derive a local `currentTopics` when `*` is active instead of reassigning the prop

alexbouchardd:
I do want to spend some time on the UI, let's hold a merge until then please 🙏

alexbouchardd:
@alexluong @Abdulmumin1 I've commited https://github.com/hookdeck/outpost/pull/912/changes/d8e5b04232f34cca8fe3e1f0202fdc2fd512cf99 following my review.

The big change is introducing `TOPICS_ALLOW_WILDCARDS`. I'm concerned that wildcard support isn't a common behaviour, and decided to make it opt-in.

Also made some small UI tweaks to the topic picker

Screenshot 2026-06-22 at 4 47 05 PM
Screenshot 2026-06-22 at 4 47 08 PM


@alexluong Ready for final review and good to merge if approved

alexluong (approved):
Great, just added a small internal change, all good from my end!

alexluong:
Thanks @Abdulmumin1 for the contribution, appreciated it!

PR #915, opened 2026-05-27
feat(portal): mobile-responsive layout

lovlyx:
Add responsive styles and small-viewport behavior across the portal UI. Key changes:

- app.scss: responsive paddings, loading margins, header layout and small-viewport positioning.
- global.scss: add $BREAKPOINT_* variables, prevent iOS zoom on inputs, hide horizontal overflow.
- Sidebar: overlay behavior below 1024px, backdrop element, mobile full-width sidebar and z-index (SCSS + Sidebar.tsx portal markup).
- Table: enable horizontal scrolling on narrow screens and set CSS var --table-min-width computed in Table.tsx (adds fallback column width calculation).
- MetricsBreakdown: constrain label widths on mobile.
- Scenes (CreateDestination, Destination, DestinationSettings, Attempts, DestinationsList): multiple layout, wrapping, grid and spacing adjustments for tablet/mobile viewports; improved wrapping for long values and tooltips.

These changes improve usability on tablet/phone widths and add safeguards for mobile browsers.

Group 427320736

### DIFF
--- a/internal/portal/src/common/Table/Table.tsx
+++ b/internal/portal/src/common/Table/Table.tsx
@@ -37,9 +37,9 @@ const Table: React.FC<TableProps> = ({
     .map((column) => (column.width ? `${column.width}px` : "1fr"))
     .join(" ");
 
-  const fallback_flex_column_width = 160;
-  const table_min_width = columns.reduce(
-    (sum, column) => sum + (column.width ?? fallback_flex_column_width),
+  const fallbackFlexColumnWidth = 160;
+  const tableMinWidth = columns.reduce(
+    (sum, column) => sum + (column.width ?? fallbackFlexColumnWidth),
     0,
   );
 
@@ -48,7 +48,7 @@ const Table: React.FC<TableProps> = ({
       className="table"
       style={
         {
-          "--table-min-width": `${table_min_width}px`,
+          "--table-min-width": `${tableMinWidth}px`,
         } as React.CSSProperties
       }
     >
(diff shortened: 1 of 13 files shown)


alexbouchardd:
Really nice! We'll do some minor tweaks and ship it 💪

internal/portal/src/common/Table/Table.tsx
```diff
@@ -37,8 +37,21 @@ const Table: React.FC<TableProps> = ({
     .map((column) => (column.width ? `${column.width}px` : "1fr"))
     .join(" ");
 
+  const fallback_flex_column_width = 160;
+  const table_min_width = columns.reduce(
+    (sum, column) => sum + (column.width ?? fallback_flex_column_width),
+    0,
+  );
```
alexluong: hey I just now noticed there's some convention issue here where we should do camelCase instead of snake_case. I get that this is following current convention, any chance you can help convert? All good if not, we can proceed with a separate refactor PR after, just want to flag this

internal/portal/src/global.scss
```diff
@@ -4,6 +4,10 @@ $BASE_GRID_MULTIPLIER: 4px !default;
 $RADIUS_S: 4px !default;
 $RADIUS_M: 6px !default;
 
+// Responsive breakpoints
+$BREAKPOINT_MOBILE: 640px !default;
+$BREAKPOINT_TABLET: 1024px !default;
```
alexluong: Doesn't seem like we're using the variables here in other CSS files?

alexluong (changes requested):

internal/portal/src/global.scss
lovlyx: Only one reference, simplified it ✅

internal/portal/src/common/Table/Table.tsx
lovlyx: updated ✅

alexbouchardd (approved):

alexluong (approved):

PR #916, opened 2026-05-28
test(spec-sdk): fix paginator response shape after Speakeasy 1.753.0

alexluong:
Speakeasy CLI bump (`1.741.7` → `1.753.0`, commit `3f311e0d`) wrapped list-operation responses in a `PageIterator`. The change wasn't flagged in the SDK regen PR changelog, so `spec-sdk-tests` silently went stale.

**Before:**
```ts
const response = await sdk.events.list({ tenantId });
const events = response.models;
```

**Now:**
```ts
// Direct access to the first page
const response = await sdk.events.list({ tenantId });
const events = response.result.models;

// Or iterate across pages
const response = await sdk.events.list({ tenantId });
for await (const page of response) {
  for (const event of page.result.models) { ... }
}

// Or step manually via the paginator
let page = await sdk.events.list({ tenantId });
while (page) {
  for (const event of page.result.models) { ... }
  page = await page.next();
}
```

This PR updates the affected tests (`events.test.ts`, `tenants.test.ts`) to the new accessor.

---

Not necessarily my personal preference (the extra `result` wrapper feels noisy for the common single-page case), but it appears to be Speakeasy convention for paginated operations — keeps the iterator metadata (`next`, `[Symbol.asyncIterator]`) co-located with the parsed body without colliding with response fields.

### DIFF
--- a/spec-sdk-tests/tests/events.test.ts
+++ b/spec-sdk-tests/tests/events.test.ts
@@ -168,7 +168,7 @@ describe('Events (PR #491)', () => {
         limit: 5,
       });
       expect(response).to.not.be.undefined;
-      expect(response?.models).to.be.an('array');
+      expect(response?.result?.models).to.be.an('array');
     });
 
     it('should list events by tenant', async function () {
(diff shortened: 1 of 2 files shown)


alexbouchardd:
Can't we change that? That seems like a totally uncecessary breaking change that actually worst then before...

alexluong:
Agreed that I don't prefer this either, will look into if there's a way to revert this behavior.

PR #945, opened 2026-06-11
docs: Clarify signature behavior

alexbouchardd:


leggetter (approved):

alexbouchardd:
@leggetter The agent eval is failing. Is it that expected to pass?

alexluong:
The agent eval is pretty flaky from what I've seen.

leggetter:
> The agent eval is pretty flaky from what I've seen.

The deterministic tests shouldn't be flaky. "LLM as a Judge" tests can be because they're based on the agent's judgment.

However, in this case the problem was with a deterministic/heuristic test caused by a change in the public signature of the TypeScript SDK. Basically a regex check.

>  - TS SDK ≤1.2.0 (e.g. commit 09b866ff): outpost.publish.event({...}) — sub-method under a publish namespace.
>  - TS SDK 1.3.0 (commit d875c662, May 13 2026, PR #905): regen flattened it to outpost.publish({...}).
 > - 1.4.0 (commit 08c270a3, PR #938): kept the flat shape.
 
So, a bug in the test that wasn't kept up to date with the SDK changes. https://github.com/hookdeck/outpost/pull/950 fixes this.

@alexluong, if you see these failing 🐛 , please do ping me. We either need to make them more reliable and therefore useful, or remove them.

PR #954, opened 2026-06-13
fix(logmq): add per-entry fallback on batch InsertMany failure

### LINKED ISSUE #948: logmq: batch insert failure nacks the entire batch — one bad entry poisons all messages
Blast-radius follow-up to #946 (escalation path) and #947 (root cause).

**Problem**

When `InsertMany` fails, `BatchProcessor.processBatch` (`internal/logmq/batchprocessor.go`) nacks every message in the batch. With the default `LOG_BATCH_SIZE=1000`, a single data-dependent failure (duplicate key as in #946, oversized row, malformed data) poisons ~1000 messages: all redeliver, the bad entry lands in a batch again, and the cycle repeats indefinitely. The mass-nack each cycle also drives ack-path API storms, which is what eventually killed the consumer in #946.

**Proposal**

On batch insert failure, fall back to per-entry inserts:

- insert each entry individually with its own short timeout
- success → continue through the normal alert-eval + ack path (factor the existing per-entry loop into a helper)
- failure → log with entry IDs, nack only that message

Properties:

- cost is only paid on the failure path; happy path unchanged
- transient errors (DB down) degrade gracefully — per-entry inserts all fail and everything nacks, same outcome as today
- data-dependent failures are isolated to the bad entries; with a DLQ/redrive policy configured, poison messages eventually leave the queue instead of dragging good ones with them

mvanhorn:
## Summary

- Add `processEntry` helper that inserts a single log entry and evaluates alerts for it
- On `InsertMany` batch failure, fall back to calling `processEntry` per-entry instead of nacking the whole batch
- Refactor post-batch alert evaluation loop into shared `evalAlert` helper (no behaviour change on happy path)
- Add two new unit tests: `TestBatchProcessor_BatchFailure_SingleBadEntry` and `TestBatchProcessor_BatchFailure_AllEntriesFail`

## Why this matters

Fixes #948. With `LOG_BATCH_SIZE=1000`, a single duplicate-key entry (SQLSTATE 21000: `ON CONFLICT DO UPDATE command cannot affect row a second time`) causes `InsertMany` to fail and every message in the batch to be nacked. All 1000 messages redeliver, the bad entry lands in a new batch, and the failure cycle repeats. The mass-nacking also exhausts the SQS consecutive-receive-error budget, permanently killing the consumer worker. The per-entry fallback limits blast radius to the one bad message; the remaining entries are acked normally. Transient failures (DB down) degrade gracefully - per-entry inserts all fail and everything nacks, same as before.

## Changes

- `internal/logmq/batchprocessor.go` - add `processEntry` and `evalAlert` helpers; call `processEntry` loop on batch insert error
- `internal/logmq/batchprocessor_test.go` - extend `mockLogStore` with `failBatchGt` field; add two batch-failure unit tests

Fixes #948

### DIFF
--- a/internal/logmq/batchprocessor.go
+++ b/internal/logmq/batchprocessor.go
@@ -136,12 +136,12 @@ func (bp *BatchProcessor) processBatch(_ string, msgs []*mqs.Message) {
 
 	insertStart := time.Now()
 	if err := bp.logStore.InsertMany(insertCtx, entries); err != nil {
-		logger.Error("failed to insert log entries",
+		logger.Error("failed to insert log entries, falling back to per-entry inserts",
 			zap.Error(err),
 			zap.Int("entry_count", len(entries)),
 			zap.Int64("insert_duration_ms", time.Since(insertStart).Milliseconds()))
-		for _, msg := range validMsgs {
-			msg.Nack()
+		for i, entry := range entries {
+			bp.processEntry(entry, validMsgs[i])
 		}
 		return
 	}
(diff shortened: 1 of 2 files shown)


alexbouchardd:
Seems to overlap with https://github.com/hookdeck/outpost/pull/953, @alexluong let's find what the right common ground between those 2 PRs is!

alexluong:
Thanks for the PR @mvanhorn, I'm review it now!

@alexbouchardd not quite overlap I think. The 2 PRs implemented 2 different GH issues. Only overlap is that it's related to the logmq flow but each PR solves different problem.

alexluong:
Hi @mvanhorn, thanks for digging into this. After some internal discussions, we're going to close this PR as well as the underlying issue #948. After some further consideration, we realized that the per-entry insertion is not the right approach for scale and we'll need to tackle this problem in a different way.

That said, the PR itself did a good job of the implementation and I appreciate the contribution! Thanks

PR #975, opened 2026-06-24
feat(webhook): cap stored response body size

### LINKED ISSUE #974: webhook: configurable max response body size to avoid oversized log messages
## Problem

A single destination response can be large enough that the resulting `LogEntry` exceeds the log queue's per-message size limit. When that happens the publish fails, the attempt never lands in the logstore, and the event retries forever (`no prior attempt found in logstore, will retry`).

The webhook driver reads the response body with `io.ReadAll` and stores it verbatim — there is no cap:

- `internal/destregistry/providers/destwebhook/httphelper.go:172` — `ParseHTTPResponse` → `io.ReadAll(resp.Body)` → `delivery.Response["body"]`
- `internal/destregistry/registry.go:205,261` — `attempt.ResponseData = deliveryData.Response`
- `LogEntry { Event, Attempt, Destination }` → `json.Marshal` → published to the log MQ

SQS limits are **256 KB per message** and 1 MB per batch. #845 capped *batch* aggregation (`MaxBatchByteSize` in `internal/mqs/queue_awssqs.go`), but a single oversized `LogEntry` still exceeds the per-message limit and is rejected — so #845 does not cover this case. It is also an unbounded-memory read regardless of queue backend.

Reported in #663 and [PR #845 discussion](https://github.com/hookdeck/outpost/pull/845#discussion_r3118229578).

## Proposal

Add a configurable max response body size for the webhook driver.

- **Default: no cap** — opt-in, fully backward-compatible.
- When set and the response body exceeds the limit, **replace the body** with a placeholder instead of storing it, e.g. `Response larger than <limit> not stored`, and flag it (`response_truncated: true` plus the size we observed). Do not store partial content.
- Enforce at **read time** in `ParseHTTPResponse` via `io.LimitReader`, so we never buffer the full body. Read up to `limit + 1`; if the extra byte is present the response is over the limit. (Note: because we stop reading early we can only report `> limit`, not the exact original size — acceptable trade-off for bounding memory.)
- Apply the placeholder to both the logstore record and the customer-visible attempt response so the two stay consistent.

### Config

New knob, env-configurable, unset = no cap. Naming TBD, e.g. `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES`.

## Scope

This exposes a knob for the webhook response body. We are deliberately not solving the general "any `LogEntry` can exceed the queue limit" problem here — operators set the limit to fit their backend.

Cases we're knowingly leaving to the operator for now:

- **Other destination types** — every provider's response funnels through `registry.go:205,261`, so a central cap would cover all of them. But only the webhook driver produces large responses (others store message IDs / ack metadata), so we're not adding a central backstop.
- **Large event payloads** — the event itself also counts toward the per-message limit. That's customer input accepted at ingest; capping it is a separate decision, not addressed here.

## Open questions

- Final config name and units (bytes vs KB).
- Placeholder wording and exact extra fields on `response_data`.

alexluong:
Bounds the destination response body stored on a delivery attempt. A large response could make the attempt's log message exceed the event queue's per-message size limit (e.g. SQS's 256 KB) — publishing then fails, the attempt never lands in the logstore, and the event retries forever (`no prior attempt found in logstore, will retry`). #845 split oversized *batches* but a single oversized message still failed; this caps the body at the source.

Opt-in via `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` (default `0` = no limit), applied to both webhook modes. Over the limit, the body is replaced with a placeholder rather than truncated, so consumers never see a partial body.

Closes #974.

**Heads-up — body is replaced, not truncated (56f8bf41):** reads via a `maxBytes+1` `io.LimitReader` to detect overflow without buffering the whole body, which means we stop before draining the response, so that connection won't be keep-alive reused — deliberate, vs. downloading an arbitrarily large body just to discard it.

### DIFF
--- a/docs/content/destinations/webhook.mdoc
+++ b/docs/content/destinations/webhook.mdoc
@@ -298,8 +298,8 @@ Configure webhook operator behavior using these keys in the Config API or in [Ho
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` | `0` (no limit) | Max bytes of the destination's response body stored on the delivery attempt (e.g. `131072` for 128 KB) |
+| `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` | `0` (no limit) | Max bytes of the destination's response body stored on the delivery attempt (e.g. `131072` for 128 KiB) |
 
-Outpost stores the destination's response body on each delivery attempt so it's visible in the event log. A very large response can make the attempt log exceed the event queue's per-message size limit — that delivery then fails to record and is retried indefinitely. Set `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` to bound the stored body — for example `131072` for a 128 KB limit. Responses larger than the limit are replaced with a placeholder (`Response body exceeded <N> bytes and was not stored`) rather than truncated, so the rest of the attempt is logged normally. The default of `0` keeps the full body.
+Outpost stores the destination's response body on each delivery attempt so it's visible in the event log. A very large response can make the attempt log exceed the event queue's per-message size limit — that delivery then fails to record and is retried indefinitely. Set `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` to bound the stored body — for example `131072` for a 128 KiB limit. Responses larger than the limit are replaced with a placeholder (`Response body exceeded <N> bytes and was not stored`) rather than truncated, so the rest of the attempt is logged normally. The default of `0` keeps the full body.
 {% /tab %}
 {% /tabs %}
--- a/docs/content/destinations/webhook.mdoc
+++ b/docs/content/destinations/webhook.mdoc
@@ -298,8 +298,8 @@ Configure webhook operator behavior using these keys in the Config API or in [Ho
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` | `0` (no limit) | Max bytes of the destination's response body stored on the delivery attempt (e.g. `131072` for 128 KiB) |
+| `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` | `131072` (128 KiB) | Max bytes of the destination's response body stored on the delivery attempt. Set to `0` to disable the cap. |
 
-Outpost stores the destination's response body on each delivery attempt so it's visible in the event log. A very large response can make the attempt log exceed the event queue's per-message size limit — that delivery then fails to record and is retried indefinitely. Set `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` to bound the stored body — for example `131072` for a 128 KiB limit. Responses larger than the limit are replaced with a placeholder (`Response body exceeded <N> bytes and was not stored`) rather than truncated, so the rest of the attempt is logged normally. The default of `0` keeps the full body.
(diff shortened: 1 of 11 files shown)


docs/content/self-hosting/configuration.mdoc
```diff
@@ -122,6 +122,7 @@ Choose one for event log persistence:
 | `DESTINATIONS_WEBHOOK_DISABLE_DEFAULT_SIGNATURE_HEADER` | `false` | Disable the signature header |
 | `DESTINATIONS_WEBHOOK_SIGNATURE_ALGORITHM` | `hmac-sha256` | Signature algorithm |
 | `DESTINATIONS_WEBHOOK_SIGNATURE_ENCODING` | `hex` | Encoding: `hex` or `base64` |
+| `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` | `0` (no limit) | Max bytes of a destination response body stored on the delivery attempt (e.g. `131072` for 128 KB). Larger responses are replaced with a placeholder so the attempt log stays under the event queue's per-message size limit. |
```
alexbouchardd: Same

docs/content/destinations/webhook.mdoc
```diff
@@ -293,5 +293,13 @@ Configure webhook operator behavior using these keys in the Config API or in [Ho
 | `DESTINATIONS_WEBHOOK_SIGNATURE_ENCODING` | `hex` | Encoding: `hex` or `base64` |
 | `DESTINATIONS_WEBHOOK_SIGNATURE_CONTENT_TEMPLATE` | `{{.Body}}` | Template for signed content in default mode |
 | `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_TEMPLATE` | `v0={{.Signatures \| join ","}}` | Template for signature header value in default mode |
+
+### Response Handling
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `DESTINATIONS_WEBHOOK_MAX_RESPONSE_BODY_BYTES` | `0` (no limit) | Max bytes of the destination's response body stored on the delivery attempt (e.g. `131072` for 128 KB) |
```
alexbouchardd: It's 12800b = 128kb, 131072 is 128 kib no?

alexbouchardd (commented):
Should we have default max values based on the queue used instead of defaulting to 0? Pubsub and SQS both have per message limits, no point in allowing more then that

alexluong:
> Should we have default max values based on the queue used instead of defaulting to 0? Pubsub and SQS both have per message limits, no point in allowing more then that

I'm not sure if that number is the right limit. Each entry log is the Event + Attempt. It has the body but also other bytes to wrap the data, so we can't be exact. But most importantly it contains the event.data which is sort of variable. That's why it could be a bit tricky to come up with a reasonable default here.

docs/content/destinations/webhook.mdoc
alexluong: updated to `KiB` instead, good catch!

alexbouchardd:
If it's tricky for us, it's even harder for the user. In this case, I'd tend to default to something fairly conservative?

The other option here which we haven't really discussed yet is to introduce compression...

alexluong:
Fair, we can consider a conservative number for this. But in this context, what if the event payload exceeds the limit? We can't _not_ store it because the data is required for retries. Should we have the same mechanism on the event publishing input validation?

alexbouchardd:
Right I do think one possibility here is that we have a default conservative response limit and the operator can tweak it depending on their max size they publish for the event data. Since it's operator controlled seems odd to have a explicit limit on that

alexluong:
@alexbouchardd I went with 128KiB which should be sufficient for most cases. Folks can configure it further if that's an issue.

alexbouchardd:
Seems reasonable, it gets expensive beyond that anyway

alexbouchardd (approved):

Issue #981, opened 2026-06-30
Emit attempt.success / attempt.failed operator events


alexluong:
Emit per-attempt operator events (`attempt.success`, `attempt.failed`) so consumers can track per-delivery results.

PR #989, opened 2026-07-03
feat(opevents): attempt.success and attempt.failed operator events

alexluong:
implements #981

Adds two operator events, plus config-based optimizations along the way.

**New topics: `attempt.success` and `attempt.failed`** — one event per delivery attempt, emitted from logmq after the attempt is persisted. The payload matches the `exhausted_retries` shape. Note: existing `*` subscribers start receiving every attempt event on upgrade — worth a release-note callout.

**Skip unnecessary work when config doesn't need it** — with alerting disabled, failed attempts skip the alert logic and its Redis calls; with attempt events also unsubscribed, entries just persist and ack.

### DIFF
--- a/docs/content/features/operator-events.mdoc
+++ b/docs/content/features/operator-events.mdoc
@@ -31,7 +31,11 @@ Available topics:
 | `alert.destination.consecutive_failure` | Consecutive failure count reaches 50%, 70%, 90%, or 100% of `ALERT_CONSECUTIVE_FAILURE_COUNT` |
 | `alert.destination.disabled` | Destination auto-disabled at 100% failure threshold |
 | `alert.attempt.exhausted_retries` | Delivery exhausts all retry attempts (deduplicated per event+destination) |
+| `attempt.success` | Every successful delivery attempt |
+| `attempt.failed` | Every failed delivery attempt, including retries |
 | `tenant.subscription.updated` | Destination created/updated/deleted and tenant topics or destination count changed |
+
+The `attempt.success` and `attempt.failed` topics fire once per delivery attempt, so they dominate event volume — including under `*`. Subscribe to them deliberately and size your sink for your delivery throughput.
 
 {% tabs tabGroup="deployment" %}
 {% tab label="Managed" %}
@@ -194,6 +198,25 @@ Emitted when a delivery exhausts all retry attempts. Deduplicated per event+dest
 }
 ```
 
+### `attempt.success` / `attempt.failed`
+
+Emitted once per delivery attempt, keyed by outcome. The two topics share one payload shape; `attempt.status` carries the outcome. `attempt.failed` fires on every failed attempt including retries — use `attempt.attempt_number` and `event.eligible_for_retry` to distinguish retryable failures from final ones.
+
+```json
+{
+  "tenant_id": "tenant_123",
+  "event": {},
+  "attempt": {},
+  "destination": {
+    "id": "des_456",
+    "tenant_id": "tenant_123",
+    "type": "webhook",
+    "topics": ["order.created"],
+    "disabled_at": null
+  }
+}
+```
+
 ### `tenant.subscription.updated`
 
 Emitted when destination changes affect tenant-level subscribed topics or destination count.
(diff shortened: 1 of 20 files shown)


docs/content/features/operator-events.mdoc
```diff
@@ -194,6 +198,25 @@ Emitted when a delivery exhausts all retry attempts. Deduplicated per event+dest
...
 
+### `attempt.success` / `attempt.failed`
+
+Emitted once per delivery attempt, keyed by outcome. The two topics share one payload shape; `attempt.status` carries the outcome. `attempt.failed` fires on every failed attempt including retries — use `attempt.attempt_number` and `event.eligible_for_retry` to distinguish retryable failures from final ones.
+
+```json
+{
+  "tenant_id": "tenant_123",
+  "event": {},
+  "attempt": {},
```
alexbouchardd: Should have example event/attempt objects here

internal/logmq/batchprocessor.go
```diff
@@ -110,6 +114,11 @@ type BatchProcessor struct {
 	alerts      AlertPipeline
 	batcher     *batcher.Batcher[*mqs.Message]
 	emitTimeout time.Duration
+	// alertsEnabled and emitsAttemptEvents are derived once from the pipeline's
+	// static config. alertsEnabled=false skips the replay gate on the failed
+	// path (nothing to protect); both false skips per-entry work entirely.
+	alertsEnabled      bool
+	emitsAttemptEvents bool
```
alexbouchardd, line 124: Not clear on the details but odd to me that this is a bool rather then list of topics. Wouldn't the behavior have to change if say the subscribed topic is `attempt.failed` but it's a success etc.

internal/logmq/batchprocessor.go
```diff
@@ -136,6 +145,11 @@ func NewBatchProcessor(ctx context.Context, logger *logging.Logger, logStore Log
 	if bp.emitTimeout <= 0 {
 		bp.emitTimeout = emitTimeout
 	}
+	if alerts.Evaluator != nil {
+		bp.alertsEnabled = alerts.Evaluator.SignalsEnabled()
+		bp.emitsAttemptEvents = alerts.Emitter.Enabled(opevents.TopicAttemptSuccess) ||
```
alexbouchardd: Also odd to me here that this is nested in the alerts.Evaluator != nil condition since the attempt events are decoupled from the alerts?

internal/logmq/batchprocessor.go
```diff
@@ -315,6 +334,29 @@ func (bp *BatchProcessor) processEntry(ctx context.Context, entry *models.LogEnt
...
+	}
+
+	// With every alert signal off there is no streak to protect and no verdict
+	// to compute, so the replay gate buys nothing — skip its two Redis round
+	// trips and emit attempt.failed ungated, same at-least-once treatment as
+	// attempt.success (a redelivery may re-emit; tolerated).
+	if !bp.alertsEnabled {
+		failed := deliveryEvent{
+			event: opevents.AttemptFailedEvent(opevents.NewAlertDestination(entry.Destination), entry.Event, entry.Attempt),
+		}
```
alexbouchardd, line 357: Same here?

internal/logmq/batchprocessor.go
alexluong: These are more of an invariant check because alerts.Evaluator always exists. I'm redoing the logic a bit so that the code doesn't look weird.

internal/logmq/batchprocessor.go
alexluong: The bools are more for logic. Whereas the routing is in the `Emitter.Emit` where it knows what are the subscribed logic.

So basically if alerts enabled -> do alert logic
If either attempts.success or attempts.failed -> send to Emitter.Emit to send the actual subscribed opevent

The batch processor logic doesn't care about what topics are subscribed. It just needs to know if it should run Evaluator.Evaluate or Emitter.Emit or not.

alexbouchardd (approved):

Issue #982, opened 2026-07-01
Support fully customizable webhook signature header name (not just prefix)


ambroziepaval:
**Is your feature request related to a problem? Please describe.**
We're planning to migrate our webhook infrastructure to Outpost. We already have thousands of integrated clients receiving webhooks from us today, and our existing contract with them specifies a fixed signature header name that does not use the -signature suffix. For example, something like x-acme-digest.

Today, `DESTINATIONS_WEBHOOK_HEADER_PREFIX` lets us customize the header prefix (e.g. x-acme-), but the suffix signature is hardcoded, so we can only ever produce x-acme-signature.

Because we already have an established contract with our clients, we can't ask them all to change the header they read. To make the migration to Outpost as seamless as possible (ideally zero client-side changes), we need to emit the signature under our existing header name.

**Describe the solution you'd like**
Make the signature header name fully customizable, not just the prefix. Something like:

A new field on DestinationWebhookConfig, e.g. `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_NAME`.
When set, this value is used as the complete signature header name, bypassing the `<prefix>signature` construction entirely. When unset, behavior is unchanged (defaults to <prefix>signature), so this is fully backwards compatible.
Example: with `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_NAME=x-acme-digest`, the delivered header would be exactly `x-acme-digest` regardless of the configured prefix.
The key point is control over the entire header name. The prefix option only covers part of the naming; consumers with an existing, fixed header contract need to pin the full name.

This mirrors the existing precedent where the signature value is already customizable via `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_TEMPLATE`: we'd simply be extending that same "make signatures interoperable with existing consumers" philosophy to the header name.

**I'm happy to implement this and open a PR.** I've traced the two call sites above and reviewed the config-field workflow in contributing/config.md (struct field + tags, defaults, LogConfigurationSummary(), validation, go generate for docs). I just wanted to align on the approach and naming before writing the code.

**Describe alternatives you've considered**
- Custom headers: We can add arbitrary static headers per destination, but those carry static values. They can't carry the computed per-request signature, so this doesn't solve our case.
- Header prefix only (`DESTINATIONS_WEBHOOK_HEADER_PREFIX`): Adjusting the prefix still leaves the mandatory signature suffix, producing x-acme-signature rather than `x-acme-digest`.
[...]

PR #992, opened 2026-07-06
feat(webhook): configurable system header names, deprecate DISABLE_* flags

ambroziepaval:
Add the global DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_NAME config option. When set, its exact value becomes the signature header name, bypassing the "<prefix>signature" construction; when unset, behavior is unchanged, so the change is fully backwards compatible.

Scoped to default mode only: standard-webhooks mode fixes the header as "webhook-signature" per spec, and renaming it would break spec-compliant verifiers. This mirrors every other signature-customization field and the existing SignatureHeaderTemplate plumbing. The name is validated at construction against the header-name regex and reserved-header set.

Issue #982.


alexbouchardd:
As mentioned in https://github.com/hookdeck/outpost/issues/982#issuecomment-4870066019, I do not believe that's the right approach, since we now have 2 configs that implement essentially the same thing. The approach should also be header-agnostic and cover every header value.

We also need to account for backward compatibility of existing deployments, possibly by keeping the `DISABLE_*` config but marking it as deprecated and printing a warning.

ambroziepaval:
Thanks @alexbouchardd , this makes sense and I want to confirm I've got your point right. The signature header already has two configs that overlap: `DISABLE_DEFAULT_SIGNATURE_HEADER` (emit or not) and `SIGNATURE_HEADER_TEMPLATE` (the value). "Disable" and "empty value" are the same outcome, and my new `SIGNATURE_HEADER_NAME` piles a third knob onto the same header.

**Concrete step I'd take now**: keep the two configs that do distinct things: `SIGNATURE_HEADER_NAME` (the header name) and `SIGNATURE_HEADER_TEMPLATE` (the value), and treat an empty template (SIGNATURE_HEADER_TEMPLATE='') as "no signature emitted". So we go from three signature knobs down to two, with no overlap. `DISABLE_DEFAULT_SIGNATURE_HEADER` stays working but deprecated with a warning for existing deployments. That directly removes the config creep you flagged.

**One scope question**: Should this stay signature-only, or give the same treatment to _event-id_, _timestamp_, and _topic_ too (deprecating their `DISABLE_DEFAULT_*_HEADER` flags the same way)?

alexluong:
Hi @ambroziepaval, thanks for the PR and sorry for the miscommunication/confusion. It was caused by me misunderstanding what @alexbouchardd meant. We connected internally and agreed on the design [here](https://github.com/hookdeck/outpost/issues/982#issuecomment-4894947752). Can you take a look and update the PR accordingly? We can chat more in the issue thread if there's anything you want to discuss further.

internal/config/destinations.go
```diff
@@ -83,20 +97,67 @@ func (c *DestinationWebhookConfig) toConfig() *destregistrydefault.DestWebhookCo
...
+		SignatureAlgorithm:       c.SignatureAlgorithm,
+		SigningSecretTemplate:    c.SigningSecretTemplate,
+		MaxResponseBodyBytes:     c.MaxResponseBodyBytes,
+	}
+}
+
+// resolveWebhookHeaderName applies the three-state rule for a webhook header
+// name, folding in the deprecated DISABLE_* flag for backward compatibility:
+//   - name set to a value  -> {Name: value}                 (pin exact name)
+//   - name set to ""       -> {Disabled: true}              (disable)
```
alexluong: I think we have a scenario of value being a whitespace of some sort `" "` that we should treat as disabled as well.

For now, as I'm tracing it, it will be treated as default headers instead.

internal/config/config_test.go
```diff
@@ -455,6 +457,151 @@ destinations:
 	}
 }
 
+func TestDestinationWebhookHeaderNames(t *testing.T) {
```
alexluong, line 460: let's add the `" "` case as well as mentioned above

internal/destregistry/providers/destwebhook/destwebhook.go
```diff
@@ -247,6 +265,41 @@ func New(loader metadata.MetadataLoader, basePublisherOpts []destregistry.BasePu
 		return nil, fmt.Errorf("signing secret template is required")
 	}
 
+	// Validate the four system headers: reject invalid/reserved pinned names, and
+	// reject collisions between the effective names (pinned name, or "<prefix>" +
+	// key) of two enabled headers. A collision would make emission order-dependent
+	// since Format writes headers from a map, so whichever iterated last would
+	// silently win.
```
alexluong, line 272: I wonder if this blob is better suited at Config.Validate() instead? I know this already runs on startup but this feels like config validation.

That's my intuition but I'm not super super opinionated there. I get that this is destwebhook logic so I can be talked into this as well haha 😅

docs/content/destinations/webhook.mdoc
```diff
@@ -271,18 +271,26 @@ Set `DESTINATIONS_WEBHOOK_PROXY_URL` to your proxy URL (basic auth supported). S
...
-| `DESTINATIONS_WEBHOOK_DISABLE_DEFAULT_TIMESTAMP_HEADER` | `false` | Disable the timestamp header |
-| `DESTINATIONS_WEBHOOK_DISABLE_DEFAULT_TOPIC_HEADER` | `false` | Disable the topic header |
-| `DESTINATIONS_WEBHOOK_DISABLE_DEFAULT_SIGNATURE_HEADER` | `false` | Disable the signature header |
+| `DESTINATIONS_WEBHOOK_EVENT_ID_HEADER_NAME` | — | Complete name of the event ID header. Unset uses the default `<prefix>event-id`; an explicit value pins that exact name; an empty string disables the header. |
+| `DESTINATIONS_WEBHOOK_TIMESTAMP_HEADER_NAME` | — | Complete name of the timestamp header. Unset uses the default `<prefix>timestamp`; an explicit value pins that exact name; an empty string disables the header. |
+| `DESTINATIONS_WEBHOOK_TOPIC_HEADER_NAME` | — | Complete name of the topic header. Unset uses the default `<prefix>topic`; an explicit value pins that exact name; an empty string disables the header. |
+| `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_NAME` | — | Complete name of the signature header. Unset uses the default `<prefix>signature`; an explicit value pins that exact name; an empty string disables the header. |
```
alexluong, line 285: In our OpenAPI schema file, there's `ManagedConfig` that we should update to include these values as well (and probably remove the DISABLED configs)

alexluong (commented):
Great PR! Overall looks great, just got a few small comments here, but otherwise everything makes sense!

internal/destregistry/providers/destwebhook/destwebhook.go
ambroziepaval: The reason I kept it hereis that the two meaningful checks both depend on things in the destwebhook package: the invalid/reserved name check needs headerNameRegex and reservedHeaders, and the collision check needs resolveHeaderName to know what each header actually resolves to. Moving it to Config.Validate() would mean either duplicating those definitions or introducing a config → destwebhook dependency. 

Happy to move it if you feel strongly though! 😅

internal/config/config_test.go
ambroziepaval: Added a test case for this

internal/config/destinations.go
ambroziepaval: Fixed

docs/content/destinations/webhook.mdoc
ambroziepaval: Done

alexluong (approved):

alexluong:
Thanks @ambroziepaval for the contribution! LGTM and I'll kick off a new release with this soon!

PR #994, opened 2026-07-13
fix(api): normalize destination topics on write

### LINKED ISSUE #990: Destination topics: duplicates and redundant wildcard states are persisted
**Affects:** `Topics` validation/storage (`internal/models`), destination create/update API; portal `TopicPicker` in wildcard mode. Relevant since #912 (`TOPICS_ALLOW_WILDCARDS`).

## Backend: the topics array is unnormalized data

`Topics.Validate` checks each entry independently — is it an available topic, or a pattern matching ≥1 available topic. It never looks at the set as a whole, so all of these are accepted and persisted verbatim:

```jsonc
["user.created", "user.created"]          // exact duplicates
["user.*", "user.created"]                // wildcard + entry it covers
["*.created", "user.created", "user.*"]   // overlapping patterns + covered entry
```

Delivery is unaffected (matching is any-entry-wins), but the stored array is ambiguous as *data*:

- Duplicates come back verbatim on every read — API responses, SDK results, portal display, `tenant.subscriptions.updated` operator event payloads.
- There's no defined semantics for what a redundant set means for **editing**: if `["user.*", "user.created"]` is stored and the wildcard is later removed, is the destination still subscribed to `user.created`? Today the answer depends on which client edits it and how.
- Since #912 these states aren't just API-reachable — the portal creates them (see below), so redundant sets will occur organically in production data.

## Portal: wildcard editing corrupts subscriptions in these states

Setup: `TOPICS=user.created,user.updated,user.deleted`, wildcards enabled.

The portal creates the mixed state itself: check `user.created`, then add `user.*` via the search field — the now-covered explicit entry is left in place, giving `["user.*", "user.created"]`.

From there, **unchecking a single topic while its group wildcard is selected** (which converts the wildcard into explicit topics) goes wrong three ways:

| | Start state | Action | Portal submits | Expected |
|---|---|---|---|---|
| 1. Writes duplicates | `["user.*", "user.created"]` | uncheck `user.updated` | `["user.created", "user.created", "user.deleted"]` | `["user.created", "user.deleted"]` |
| 2. Can't unsubscribe | `["user.*", "user.created"]` | uncheck `user.created` | `["user.created", "user.updated", "user.deleted"]` — checkbox snaps back to checked | `["user.updated", "user.deleted"]` |
| 3. Filter wipes hidden topics | `["user.*"]` | search `created`, uncheck `user.created` | `[]` — one unchecked box empties the subscription | `["user.updated", "user.deleted"]` |

All three live in the same code path: the `hasWildcard` branch of the child checkbox `onChange` (`TopicPicker.tsx`, ~line 329). It removes only the exact `wildcardId` from the current selection (rows 1–2: covered entries survive the conversion) and rebuilds the explicit list from the search-filtered `visibleTopics` (row 3).

## Suggested direction

- **Normalize on write**: dedupe exact duplicates and fold entries covered by a sibling pattern (`["user.*", "user.created"]` → `["user.*"]`). One consideration for the folding case: it discards the explicit entry, so removing `user.*` later also removes the `user.created` subscription — acceptable, but worth being deliberate about.
- **Overlapping patterns** (`["*.created", "user.*"]`) are a state to be aware of: neither covers the other, so normalization leaves both — pattern-vs-pattern redundancy is out of scope here.
- The portal conversion needs fixing regardless: redundant states are already persistable, so the editor must handle encountering them.

mvanhorn:
## Summary

`Topics.Validate` checks each entry independently, so a destination's stored
topic set can contain redundant entries that are accepted verbatim:

- exact duplicates: `["user.created", "user.created"]`
- a wildcard plus an entry it covers: `["user.*", "user.created"]`
- overlapping patterns plus a covered entry: `["*.created", "user.created", "user.*"]`

Delivery is unaffected (topic matching is any-entry-wins), but the persisted
array is ambiguous as data: duplicates echo back on every read (API/SDK/portal
and `tenant.subscriptions.updated` payloads), and there is no defined edit
semantics for a redundant set. Since #912 (`TOPICS_ALLOW_WILDCARDS`) the portal
itself can create these mixed states, so they occur organically.

This adds `Topics.Normalize()` and calls it after `Validate` in the destination
create and update handlers:

- dedupes exact duplicates, keeping the first occurrence;
- folds an entry that a sibling wildcard pattern strictly covers
  (`["user.*", "user.created"]` -> `["user.*"]`), where "strictly covers" means
  the pattern covers the entry but the entry does not cover the pattern;
- leaves mutually-non-covering pattern pairs untouched
  (`["*.created", "user.*"]` is unchanged);
- leaves `["*"]` and already-minimal sets untouched;
- preserves first-seen order so API responses stay stable.

Because it only removes entries already covered by a retained entry, `MatchTopic`
(delivery) results are identical before and after. A create/update with
`["user.*", "user.created"]` now stores and returns `["user.*"]`.

## Why this matters

This implements the "normalize on write" direction from the issue author
(@alexluong): "dedupe exact duplicates and fold entries covered by a sibling
pattern (`["user.*", "user.created"]` -> `["user.*"]`)", reaffirmed as "if
`user.*` then should not have `user.created`". Normalizing at the write boundary
makes the persisted topic set unambiguous, so every downstream reader (API, SDK,
portal, subscription-update events) sees a single canonical representation. It
also gives the portal topic editor a well-defined data model to build on.

### Out of scope (deliberately)

- Portal `TopicPicker.tsx` `hasWildcard` `onChange` fixes. That is a separate
  frontend change requiring the portal running to verify; this backend
  normalization is the prerequisite that makes the persisted data unambiguous.
- Rejecting non-trailing wildcards such as `*.created` ("wildcard as last param",
  raised by @alexbouchardd). That is a validation-tightening change that could
  reject currently-valid stored data, a distinct decision from normalization -
  left for a maintainer as a known follow-up.

Per the issue's own note, pattern-vs-pattern overlap where neither covers the
other (`["*.created", "user.*"]`) is intentionally left intact.

## Tests

- New `TestDestinationTopics_Normalize` table test covering dedupe, fold,
  mutually-non-covering patterns, `["*"]`, already-minimal sets, and order
  preservation.
- Existing `TestDestinationTopics_Validate` unchanged.
- `go build ./...`, `go test ./internal/models/... ./internal/apirouter/...`,
  and `go vet` on the changed packages all pass.

Closes #990

### DIFF
--- a/internal/apirouter/destination_handlers.go
+++ b/internal/apirouter/destination_handlers.go
@@ -109,6 +109,7 @@ func (h *DestinationHandlers) Create(c *gin.Context) {
 		AbortWithValidationError(c, err)
 		return
 	}
+	destination.Topics = destination.Topics.Normalize()
 	if err := h.registry.ValidateDestination(c.Request.Context(), &destination); err != nil {
 		AbortWithValidationError(c, err)
 		return
@@ -183,6 +184,7 @@ func (h *DestinationHandlers) Update(c *gin.Context) {
 			AbortWithValidationError(c, err)
 			return
 		}
+		updatedDestination.Topics = updatedDestination.Topics.Normalize()
 	}
 	shouldRevalidate := false
 	if input.Type != "" && input.Type != originalDestination.Type {
--- a/internal/apirouter/destination_handlers_test.go
+++ b/internal/apirouter/destination_handlers_test.go
@@ -142,6 +142,46 @@ func TestAPI_Destinations(t *testing.T) {
 			require.Equal(t, http.StatusUnprocessableEntity, resp.Code)
 		})
 
+		t.Run("duplicate topics are normalized", func(t *testing.T) {
+			h := newAPITest(t)
+			h.tenantStore.UpsertTenant(t.Context(), tf.Any(tf.WithID("t1")))
+
+			payload := validDestination()
+			payload["topics"] = []string{"user.created", "user.created"}
+
+			req := h.jsonReq(http.MethodPost, "/api/v1/tenants/t1/destinations", payload)
+			resp := h.do(h.withAPIKey(req))
+
+			require.Equal(t, http.StatusCreated, resp.Code)
+			var dest destregistry.DestinationDisplay
+			require.NoError(t, json.Unmarshal(resp.Body.Bytes(), &dest))
+			assert.Equal(t, models.Topics{"user.created"}, dest.Topics)
+
+			stored, err := h.tenantStore.RetrieveDestination(t.Context(), "t1", dest.ID)
+			require.NoError(t, err)
+			assert.Equal(t, models.Topics{"user.created"}, stored.Topics)
+		})
+
+		t.Run("topic covered by sibling wildcard is normalized", func(t *testing.T) {
+			h := newAPITest(t, withTopicsAllowWildcards(true))
+			h.tenantStore.UpsertTenant(t.Context(), tf.Any(tf.WithID("t1")))
+
+			payload := validDestination()
+			payload["topics"] = []string{"user.*", "user.created"}
+
+			req := h.jsonReq(http.MethodPost, "/api/v1/tenants/t1/destinations", payload)
+			resp := h.do(h.withAPIKey(req))
+
+			require.Equal(t, http.StatusCreated, resp.Code)
+			var dest destregistry.DestinationDisplay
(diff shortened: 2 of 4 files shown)


alexluong (approved):
Thanks for the PR, LGTM!

One small comment is can we add some tests at the handler level where we send POST/PUT/PATCH request with different topics input and confirm that the topic list is normalized?

Issue #463, opened 2025-08-21
Feature Request: Add support for AWS IAM roles (ECS Task Role / EC2 Instance Profile) for SQS authentication


moltar:
# Feature Request: Add support for AWS IAM roles (ECS Task Role / EC2 Instance Profile) for SQS authentication

## Description

Currently, Outpost requires explicit AWS credentials (`AWS_SQS_ACCESS_KEY_ID` and `AWS_SQS_SECRET_ACCESS_KEY`) for SQS authentication. This prevents the use of AWS IAM roles, which is the recommended security practice for applications running on AWS infrastructure like ECS, EC2, or Lambda.

## Current Behavior

When running Outpost on ECS Fargate without explicit credentials, the service fails with:
```
config validation error: message queue configuration is required
```

Looking at the source code in `internal/config/mqconfig_aws.go`, the `IsConfigured()` method requires non-empty credentials:

```go
func (c *AWSSQSConfig) IsConfigured() bool {
    return c.AccessKeyID != "" && c.SecretAccessKey != "" && c.Region != ""
}
```

## Expected Behavior

Outpost should support AWS IAM roles by:
1. Detecting when running in an AWS environment with IAM roles available
2. Using the AWS SDK's default credential provider chain
3. Only requiring the region when IAM roles are available

## Use Case

Many organizations follow AWS security best practices by:
- Using IAM roles instead of long-lived access keys
- Running containerized workloads on ECS with task roles
- Deploying on EC2 instances with instance profiles
- Following the principle of least privilege

Currently, we have to create IAM users with access keys specifically for Outpost, which:
- Goes against security best practices
- Requires additional secret management
- Increases the attack surface
- Makes credential rotation more complex

## Proposed Solution

Modify the `IsConfigured()` method to support IAM roles:

```go
func (c *AWSSQSConfig) IsConfigured() bool {
    // Check if running with IAM role credentials
    if hasIAMRoleCredentials() {
        return c.Region != ""
    }
    // Fall back to explicit credentials
    return c.AccessKeyID != "" && c.SecretAccessKey != "" && c.Region != ""
}

func hasIAMRoleCredentials() bool {
    // ECS Task Role
    if os.Getenv("AWS_CONTAINER_CREDENTIALS_RELATIVE_URI") != "" {
        return true
    }
    // EC2 Instance Profile
    if os.Getenv("AWS_CONTAINER_CREDENTIALS_FULL_URI") != "" {
        return true
    }
[...]

PR #996, opened 2026-07-17
feat(mqconfig): support IAM role auth for SQS MQ provider

ambroziepaval:
The AWS SQS message-queue provider previously required static IAM user keys (AWS_SQS_ACCESS_KEY_ID / AWS_SQS_SECRET_ACCESS_KEY) and always passed them explicitly to the SDK, which suppressed the default credential chain. This made role-based auth (ECS/EKS task role, EC2 instance profile) impossible.

When no static credentials are configured, ToCredentials now returns (nil, nil) and the client builders omit WithCredentialsProvider, letting the AWS SDK default credential chain resolve credentials. IsConfigured selects SQS on region alone so the keys can be omitted. Existing static-key configs are unchanged. Mirrors the GCP Pub/Sub provider's implicit-credential behavior.

Refs hookdeck/outpost#463

### DIFF
--- a/internal/config/mqconfig_aws.go
+++ b/internal/config/mqconfig_aws.go
@@ -2,12 +2,16 @@ package config
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/hookdeck/outpost/internal/mqinfra"
 	"github.com/hookdeck/outpost/internal/mqs"
 )
+
+// errPartialAWSSQSCredentials rejects a config with exactly one static key set.
+var errPartialAWSSQSCredentials = errors.New("AWS SQS: both access_key_id and secret_access_key must be set together, or both omitted to use the default credential chain")
 
 type AWSSQSConfig struct {
 	AccessKeyID     string `yaml:"access_key_id" env:"AWS_SQS_ACCESS_KEY_ID" desc:"AWS Access Key ID for SQS. Optional: omit (with the secret access key) to use the AWS SDK default credential chain, e.g. an IAM role." required:"N"`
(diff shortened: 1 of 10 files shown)


internal/config/mqconfig_aws.go
```diff
@@ -60,6 +60,8 @@ func (c *AWSSQSConfig) GetProviderType() string {
 	return "awssqs"
 }
 
+// IsConfigured selects SQS on Region alone; keys are optional so IAM-role auth
+// can use the AWS SDK default credential chain.
 func (c *AWSSQSConfig) IsConfigured() bool {
-	return c.AccessKeyID != "" && c.SecretAccessKey != "" && c.Region != ""
+	return c.Region != ""
 }
```
alexluong, line 74: While this is correct, I discovered a config data issue where if only 1 of AWS_SQS_ACCESS_KEY_ID or AWS_SQS_SECRET_ACCESS_KEY is provided, it is accepted and led to a runtime auth error at a later stage even though this is invalid.

I don't think `IsConfigured()` is the right place but maybe we can apply this validation check at `ToQueueConfig()`?

alexluong (commented):
Thanks! Overall LGTM with a small comment.

There's also `publishmq` which has a different AWS SDK path. I think it's probably relevant and something we can apply similar logic. Not a blocker but can you include that in this PR as well?

internal/config/mqconfig_aws.go
ambroziepaval: Thanks @alexluong! addressed both points:
- added the partial-credential guard in `ToQueueConfig()` as suggested so it fails at startup via `validateMQs` instead of deferring to a runtime error
- extended IAM-role support to publishmq. The gap was in the config layer, since it already shared the IAM-ready SDK path. Relaxed the key requirementss, make selection region-only and applied the same partial-credentials check

alexluong (approved):

PR #997, opened 2026-07-20
feat(deliverymq): bound retry task receives with backoff and DLQ

LendritIbrahimi:
This fixes the issue where a single lost attempt log produced an unbounded 30s warning loop that couldn't be turned off by destination disable or pod restarts. (#663)

The implementation follows the suggestion in the linked issue, failed retry tasks now back off exponentially using rsmq's receive count, instead of reappearing every 30s forever. Tasks received more than RETRY_MAX_RECEIVE_COUNT times are moved to the deliverymq-retry-dlq queue. 

The move to the DLQ is send-then-delete, so a failure between the two steps can only duplicate a task, never lose one, and landing there emits an error log with the full task payload for operator visibility. RETRY_MAX_RECEIVE_COUNT defaults to 5 and can be set to 0 to disable the DLQ path entirely.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -82,7 +82,6 @@ type Config struct {
 	RetryMaxLimit                 int   `yaml:"retry_max_limit" env:"MAX_RETRY_LIMIT" desc:"Maximum number of retry attempts for a single event delivery before giving up. Ignored if retry_schedule is provided." required:"N"`
 	RetryPollBackoffMs            int   `yaml:"retry_poll_backoff_ms" env:"RETRY_POLL_BACKOFF_MS" desc:"Backoff time in milliseconds when the retry monitor finds no messages to process. When a retry message is found, the monitor immediately polls for the next message without delay. Lower values provide faster retry processing but increase Redis load. For serverless Redis providers (Upstash, ElastiCache Serverless), consider increasing to 5000-10000ms to reduce costs. Default: 100" required:"N"`
 	RetryVisibilityTimeoutSeconds int   `yaml:"retry_visibility_timeout_seconds" env:"RETRY_VISIBILITY_TIMEOUT_SECONDS" desc:"Time in seconds a retry message is hidden after being received before becoming visible again for reprocessing. This applies when event data is temporarily unavailable (e.g., race condition with log persistence). Default: 30" required:"N"`
-	RetryMaxReceiveCount          int   `yaml:"retry_max_receive_count" env:"RETRY_MAX_RECEIVE_COUNT" desc:"Maximum number of times a retry task can be received before it is moved to the deliverymq-retry-dlq queue. This is separate from retry_max_limit, which bounds delivery attempts to the destination. Set to 0 to disable. Default: 5" required:"N"`
 
 	// Event Delivery
 	MaxDestinationsPerTenant int `yaml:"max_destinations_per_tenant" env:"MAX_DESTINATIONS_PER_TENANT" desc:"Maximum number of destinations allowed per tenant/organization." required:"N"`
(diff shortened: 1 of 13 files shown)


internal/config/config.go
```diff
@@ -82,6 +82,7 @@ type Config struct {
...
 	RetryVisibilityTimeoutSeconds int   `yaml:"retry_visibility_timeout_seconds" env:"RETRY_VISIBILITY_TIMEOUT_SECONDS" desc:"Time in seconds a retry message is hidden after being received before becoming visible again for reprocessing. This applies when event data is temporarily unavailable (e.g., race condition with log persistence). Default: 30" required:"N"`
+	RetryMaxReceiveCount          int   `yaml:"retry_max_receive_count" env:"RETRY_MAX_RECEIVE_COUNT" desc:"Maximum number of times a retry task can be received before it is moved to the deliverymq-retry-dlq queue. This is separate from retry_max_limit, which bounds delivery attempts to the destination. Set to 0 to disable. Default: 5" required:"N"`
```
alexluong: I think we should hard-code these instead of making them configurable.

I think the RetryVisibilityTimeoutSeconds was an oversight. We can deal with them later. But for this PR, we can hard-code the value for RetryMaxReceiveCount in the retrymq construction. The reason is that this is an Outpost-managed queue and I think we don't need to make it configurable prematurely unless there's a reason it's necessary.

I mainly want to make it consistent with deliverymq and logmq. We don't expose knobs for retry policy for those either, so we shouldn't do so for retrymq.

From my end, 5 seems like a reasonable number to start or match whatever number we used for  deliverymq.

internal/scheduler/scheduler.go
```diff
@@ -167,9 +193,16 @@ func (s *schedulerImpl) Monitor(ctx context.Context) error {
 				time.Sleep(s.config.pollBackoff)
 				continue
 			}
+			if s.config.maxReceiveCount > 0 && msg.Rc > s.config.maxReceiveCount {
+				s.moveToDLQ(ctx, msg)
+				continue
+			}
```
alexluong, line 199: One edge case to handle before merge: the receive count can survive a `Schedule` overwrite, which makes this check fire early for a freshly scheduled task.

`Rc` lives in the `<id>:rc` hash field, incremented by the receive script and cleared only by `DeleteMessage`. But `SendMessage` with a custom ID — which is exactly what `Schedule` does with the deterministic retry ID (`eventID:destinationID`) to guarantee at most one pending retry per event+destination — overwrites the message body and score while leaving `<id>:rc` untouched.

So this sequence inherits a stale count:

1. A retry task is cycling on exec failure (rc climbing — the #663 scenario).
2. While it's still queued, a new delivery attempt for the same event+destination fails and schedules over it (e.g. a user fires a manual retry mid-incident).
3. The overwritten task is semantically brand new, but starts life with the predecessor's rc — and gets dead-lettered after fewer than `maxReceiveCount` real receives. (Same thing if exec succeeded but `DeleteMessage` failed, leaving rc behind.)

The consequence is bounded (early DLQ, never a loop or loss), but the count should mean "receives of *this* task," not accumulated history of the ID slot.

Suggested fix: in `SendMessage` (rsmq.go), when a custom ID is provided, clear the counters in the existing tx so the reset is atomic with the overwrite — mirroring what `DeleteMessage` already deletes:

```go
if options.id != "" {
    tx.HDel(key+q, messageID+":rc", messageID+":fr")
}
```

A test that schedules over a message with rc > 0 and asserts the full `maxReceiveCount` budget is available again would lock it in.

alexluong (commented):
Great work, thanks for opening the PR!

Small other nit, can you try `gofmt` to make sure everything is formatted.

internal/config/config.go
LendritIbrahimi: Done.

Small note the "0 disables the DLQ" option went away with it, so the bound is always on now

LendritIbrahimi:
Added a commit for the changes, tested E2E locally as well, if everything looks right feel free to merge 🫡

alexluong (approved):
LGTM!

PR #1004, opened 2026-07-23
docs: add webhook signature verification guide

alexluong:
Adds `guides/verify-webhook-signatures`: how default-mode webhook signatures are constructed (hex HMAC-SHA256 over the raw body, secret used as-is including the `whsec_` prefix), verification steps, and rotation-safe code examples in Node.js, Python, and Go. The only customization it accounts for is the header prefix; signature template/algorithm/encoding changes are noted as out of scope. Linked from the webhook destination page and added to the Guides nav.

The code examples were tested against real deliveries: fixtures generated through `destwebhook`'s publisher (single secret, rotation, custom prefix, unicode body) all verify with the samples verbatim, and tampered-body/wrong-secret/stripped-`whsec_` cases correctly fail.

**Heads-up:** the destination page's verification steps drop "reject old timestamps" — the default signature covers only the body, so the timestamp header isn't authenticated; the guide points to event-id dedupe instead.

### DIFF
--- a/docs/content/guides/verify-webhook-signatures.mdoc
+++ b/docs/content/guides/verify-webhook-signatures.mdoc
@@ -1,0 +1,201 @@
+---
+title: "Verify Webhook Signatures"
+description: "Verify the HMAC signature on webhooks delivered by Outpost's default webhook mode."
+---
+
+Outpost signs every webhook it delivers. Verify the signature before trusting a request — it proves the request came from the sender that holds the signing secret and that the body was not modified in transit.
+
+This guide covers Outpost's **default** webhook mode. For [Standard Webhooks](https://www.standardwebhooks.com/) mode, verify with the official [Standard Webhooks SDKs](https://github.com/standard-webhooks/standard-webhooks/tree/main/libraries) instead.
+
+## Request format
+
+Each webhook is an HTTP POST. The body is the event payload as JSON, and system headers carry the delivery metadata:
+
+```
+POST /webhooks HTTP/1.1
+Content-Type: application/json
+x-outpost-event-id: evt_abc123
+x-outpost-topic: user.created
+x-outpost-timestamp: 2024-06-01T08:23:36Z
+x-outpost-signature: v0=abc123def456...
+
+{"user_id": "usr_123", "email": "user@example.com"}
+```
+
+Header names are `<prefix>event-id`, `<prefix>topic`, `<prefix>timestamp`, and `<prefix>signature`. The default prefix is `x-outpost-`; deployments commonly set their own (for example `x-acme-signature`). Substitute the actual prefix throughout this guide. Header names are case-insensitive.
+
+## How the signature is computed
+
+```
+signature = hex(HMAC-SHA256(secret, body))
+```
+
+- **`secret`** is the destination's signing secret, used exactly as issued. Secrets typically look like `whsec_a1b2c3...` — the whole string, including the `whsec_` prefix, is the HMAC key. Do not strip the prefix or base64-decode it.
+- **`body`** is the raw request body bytes, exactly as received.
+- The digest is hex-encoded.
+
+The signature header value is `v0=` followed by one or more comma-separated signatures:
+
+```
+x-outpost-signature: v0=<signature>
+x-outpost-signature: v0=<signature-1>,<signature-2>
+```
+
+Multiple signatures appear during [secret rotation](#secret-rotation); the request is valid if any of them matches.
+
+## Verification steps
+
+1. Read the **raw request body** before any parsing. Deserializing and re-serializing the JSON can reorder keys or change whitespace and will break verification.
+2. Read the signature header and check it starts with `v0=`. Reject the request if the header is missing.
+3. Strip `v0=` and split the rest on `,` to get the candidate signatures.
+4. Compute `hex(HMAC-SHA256(secret, body))` with your signing secret.
+5. Compare your computed signature against each candidate using a **constant-time comparison**. Accept the request if any candidate matches; otherwise respond with `401`.
+
+## Code examples
+
+Each example uses the default `x-outpost-signature` header name — substitute the actual prefix if it differs.
+
+{% tabs tabGroup="language" %}
+{% tab label="Node.js" %}
+```javascript
+const crypto = require("crypto");
+const express = require("express");
+
+const SIGNATURE_HEADER = "x-outpost-signature";
+
+function verifySignature(rawBody, signatureHeader, secret) {
+  if (!signatureHeader || !signatureHeader.startsWith("v0=")) return false;
+  const expected = crypto
+    .createHmac("sha256", secret)
+    .update(rawBody)
+    .digest("hex");
+  return signatureHeader
+    .slice(3)
+    .split(",")
+    .some((candidate) => {
+      const a = Buffer.from(candidate);
+      const b = Buffer.from(expected);
+      return a.length === b.length && crypto.timingSafeEqual(a, b);
+    });
+}
+
+const app = express();
+
+// express.raw() keeps the body as a Buffer for verification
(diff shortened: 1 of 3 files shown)


docs/content/guides/verify-webhook-signatures.mdoc
```diff
@@ -0,0 +1,201 @@
...
+```python
+import hashlib
+import hmac
+import os
+
+from flask import Flask, request
+
+SIGNATURE_HEADER = "x-outpost-signature"
+
+def verify_signature(raw_body: bytes, signature_header: str | None, secret: str) -> bool:
```
leggetter, line 110: The `str | None` union syntax requires Python 3.10+. Users on 3.8/3.9 who paste this get a `TypeError` at import time. Consider `Optional[str]` (with `from typing import Optional`) or dropping the annotation to keep the sample maximally copy-pasteable.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/content/guides/verify-webhook-signatures.mdoc
```diff
@@ -0,0 +1,201 @@
...
+
+{% tabs tabGroup="language" %}
+{% tab label="Node.js" %}
+```javascript
+const crypto = require("crypto");
+const express = require("express");
+
+const SIGNATURE_HEADER = "x-outpost-signature";
+
+function verifySignature(rawBody, signatureHeader, secret) {
```
leggetter, line 66: Edge case: with `express.raw({ type: "application/json" })`, a request sent without a matching `Content-Type` leaves `req.body` as an empty object (not a Buffer), and `crypto.createHmac(...).update({})` throws an unhandled exception instead of returning 401. A one-line guard makes the sample robust:

```suggestion
function verifySignature(rawBody, signatureHeader, secret) {
  if (!Buffer.isBuffer(rawBody)) return false;
```

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/content/guides/verify-webhook-signatures.mdoc
```diff
@@ -0,0 +1,201 @@
...
+	// handle event...
+	w.WriteHeader(http.StatusOK)
+}
+```
+{% /tab %}
+{% /tabs %}
+
+## Secret rotation
+
+When a signing secret is rotated, both the old and new secrets remain valid for a rotation window. During that window the signature header contains two comma-separated signatures — the current secret's first, the previous secret's second:
```
leggetter: "a rotation window" — worth saying how long, or linking to where it's defined. The window is governed by `previous_secret_invalid_at` (documented in the credentials table on the [webhook destination page](/docs/outpost/destinations/webhook#configuration)), with a 24-hour fallback when unset (`internal/destregistry/providers/destwebhook/signature.go`, `GenerateSignatures`). A link would let readers find the actual duration for their deployment.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/content/guides/verify-webhook-signatures.mdoc
```diff
@@ -0,0 +1,201 @@
...
+
+## Duplicate deliveries
+
+Delivery is at-least-once — the same event can be delivered more than once, for example after a retry. The `<prefix>event-id` header carries a stable id for the event; store processed ids and skip any already handled.
+
+The default signature covers only the request body, so the timestamp and event id headers are not authenticated. Checking the timestamp's freshness adds no replay protection, and event id deduplication handles retry duplicates rather than deliberate replay. Protection against deliberate replay requires the timestamp inside the signed content — a [custom signature template](/docs/outpost/destinations/webhook#custom-signature-templates) or Standard Webhooks mode.
+
+## Customization
+
+The scheme above is Outpost's default. Changing `DESTINATIONS_WEBHOOK_HEADER_PREFIX` only renames the headers — this guide stays accurate with the prefix substituted. Customizing the [signature templates, algorithm, or encoding](/docs/outpost/destinations/webhook#signatures) changes the scheme itself, and the resulting deliveries no longer match this guide.
```
leggetter: Minor completeness point: besides the prefix, individual system header names can also be pinned to a custom name or disabled entirely (e.g. `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_NAME` — the destination page documents this for `event-id`). A brief mention that per-header overrides exist would cover deployments where prefix substitution alone doesn't explain the header names a reader sees.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/content/nav.json
```diff
@@ -82,6 +82,10 @@
             "slug": "guides/migrate-to-outpost",
             "title": "Migrate to Outpost"
           },
+          {
+            "slug": "guides/verify-webhook-signatures",
```
leggetter, line 86: Placement is fine as-is, but this is arguably the first guide most event consumers need — worth considering a higher slot in the Guides list.

---
_Generated by [Claude Code](https://claude.ai/code)_

leggetter (commented):
<details>
<summary>Review summary — accurate guide, five minor suggestions inline</summary>

I verified the guide's claims against the signing implementation and they all hold:

- Defaults match `internal/destregistry/providers/destwebhook/destwebhook.go`: hex HMAC-SHA256, content template `{{.Body}}` (body only), header template `v0={{.Signatures | join ","}}`, prefix `x-outpost-`, secret template `whsec_{{.RandomHex}}`.
- The secret is used verbatim as the HMAC key (`signature.go`), so the "don't strip `whsec_` or base64-decode" warning is correct and valuable.
- Rotation ordering (current secret's signature first) and the comma-joined multi-signature format match `GenerateSignatures`.
- The timestamp header is RFC 3339 UTC, matching the example.
- Removing "reject old timestamps" from the destination page is a genuine correctness fix — the default signed content is the body only, so the timestamp header is unauthenticated and freshness checks add no replay protection.

The code examples use constant-time comparison in all three languages, read the raw body before parsing, and handle multi-signature headers correctly.

The inline comments are all minor polish: a Python 3.10+ type-hint compatibility nit, a robustness guard for the Node example, linking the rotation-window duration, mentioning per-header name overrides in the Customization section, and an optional nav-placement thought. None are blockers.

</details>

---
_Generated by [Claude Code](https://claude.ai/code)_

leggetter (commented):
Reviewed with Claude, reviewed the review, and then got Claude to post some comments.

docs/content/guides/verify-webhook-signatures.mdoc
alexluong: Reworded to be a bit more generic instead of listing all available configuration options.

https://github.com/hookdeck/outpost/commit/89896d14c7d7f574e99b72280804b58bd6ccd222

docs/content/guides/verify-webhook-signatures.mdoc
alexluong: https://github.com/hookdeck/outpost/commit/09080f27243b0d52aac40c50eaf181ab9a39877a

done!

docs/content/guides/verify-webhook-signatures.mdoc
alexluong: https://github.com/hookdeck/outpost/commit/613903bdb4cb100a4c19e549ad5edcecb7c70cae

docs/content/guides/verify-webhook-signatures.mdoc
alexluong: Both 3.8/3.9 have hit EOL by the end of last year. I assume folks can do their own translation if necessary? Not super opinionated tho, just a thought. What do you think?

docs/content/nav.json
alexluong: I think the "Migrate to Outpost" is a more important guide. This is the 2nd one already. I personally think it makes sense this way.

alexbouchardd (approved):

PR #1021, opened 2026-08-07
fix: share one HTTP client per provider and size its connection pool

### LINKED ISSUE #1017: Webhook deliveries open a new connection for nearly every request
Outpost builds a separate HTTP client for every destination and never configures connection pooling on it.

**Each client keeps only 2 idle connections.** That's Go's default, inherited because nothing sets it. Above two concurrent deliveries to a destination, connections are closed after use instead of reused.

Under sustained load, connection reuse collapses to roughly one new connection per delivery.

## Impact

The cost lands differently depending on how fast the destination responds, and both ends are bad:

| destination | what it costs |
|---|---|
| fast (tens of ms) | **latency** — the TCP and TLS handshake is several times the request itself |
| slow (seconds) | **ephemeral ports** — closed connections accumulate in `TIME_WAIT` on the sender |

At the slow end this is the more serious one. A single destination responding near the delivery timeout can hold several thousand ports; a handful of such destinations can exhaust the sender's ephemeral range, after which deliveries start failing at the connection level. Those failures are currently recorded as a generic network error with the underlying cause discarded, which makes them hard to attribute — #1018 addresses that.

## When it triggers

The threshold is **more than two concurrent deliveries to a single destination**, which is `rate × response time`. Roughly 100 events/s to one destination at typical response times, or far less if that destination is slow.

`DELIVERY_MAX_CONCURRENCY` defaults to 1, so a default deployment never reaches it. This affects deployments that have raised it.

## Why the per-destination client exists

No reason, on inspection. Nothing in the client configuration varies per destination — user agent, proxy settings, and the transport wrapper all come from provider-level options fixed at registration, and there's no per-destination TLS configuration anywhere in the registry. We build many structurally identical clients and give each its own small pool.

## Proposed change

**Build the client once per distinct transport configuration** — today, once per provider at registration, rather than per destination in `CreatePublisher`.

This makes the per-host idle limit meaningful and gives a real ceiling on total idle connections. There's precedent: the `hookdeck` provider already accepts an injected shared client.

The rule worth preserving: **only connection-level concerns justify a separate client.** Proxy settings and TLS material qualify. Headers, auth, timeouts and response body limits are per-request and belong on the request. If per-destination proxy or client certificates are needed later, transports should be keyed by configuration and shared within a key — not built per destination.

## Sizing the pool

Two limits matter, and they mean different things:

- **per-host** — how many warm connections a *single* destination keeps. This is depth, bounded by concurrent deliveries to that destination.
- **total** — how many destinations can keep a warm connection *at all*. This is breadth. If it's too small, destinations evict each other's connections and reuse never happens regardless of the per-host value.

Both derive from `DELIVERY_MAX_CONCURRENCY` (call it C) — sane defaults scaling with the one knob operators already set for delivery throughput:

- **per-host** = `max(C, 2)` — the worker pool caps how many deliveries can be in flight, so it caps how many connections one destination could need; 2 is Go's default, so C=1 never sizes below stock.
- **total** = `clamp(32 × C, 512, max(4096, C))`.

The total limit caps only parked (idle) connections, never active ones — exceeding it causes connection churn, not errors — so the constants are reuse-rate tuning, not resource limits. 32 × C ≈ how many distinct destinations a worker revisits within the 90s `IdleConnTimeout` at ~3s per delivery (slow destinations are where reuse matters most). The floor of 512 covers low-concurrency fanout: concurrency bounds simultaneous requests, not distinct hosts touched over time — C=1 at ~100ms per delivery still cycles ~900 destinations per idle window. The cap of 4096 bounds the parked-FD/memory cost where the hit rate decays, and rises to C so it never undersizes the pool below the concurrency level.

An earlier revision derived the total from `RLIMIT_NOFILE`. Dropped: the limit is per-process and unreliable in containers — the `RLIM_INFINITY` quirk meant the most generous environments derived the *smallest* pools — and mq-provider publishers hold per-destination FDs the derivation never accounted for. Since the limit only governs idle connections, tying it to a hard resource ceiling was solving the wrong problem.

The `hookdeck` provider should get a small fixed value instead — it talks to one host, so it needs depth, not breadth.

## No new configuration

Deliberately not exposing these as env vars yet. They are defaults that scale with a knob operators already set; a standalone knob invites confident wrong values before the reuse metric exists to judge them by.

Adding a knob later is backward-compatible; removing one isn't. If the derivation proves wrong, the escape hatch can be added then.

Instead:

- **log the resolved values at startup**
- **add a connection reuse metric** — connections opened against deliveries attempted. This is the signal that the ceiling is binding, and it's currently the thing we have no way to observe in production.

## Verification

A test asserting **connections opened stays near concurrency, not near request count**, swept across concurrency levels and against both fast and slow destinations, since the two regimes fail differently.

## Tradeoffs worth stating

**Cross-tenant pooling.** One provider serves all tenants, so the pool becomes shared. Connections are per-host so there's no correctness concern, but a busy tenant can evict a quiet tenant's idle connections. The per-host limit bounds the blast radius.

**Connections held on the destination.** Idle connections occupy a slot on the receiver's server. Those connections were opened anyway to serve the traffic; pooling changes whether they're held afterward, for up to 90 seconds. Bounded, but a real behavior change from the receiver's side.

alexluong:
Closes #1017.

Outpost built a separate `http.Client` for every destination and never configured pooling on it, so each inherited Go's default of two idle connections. Above two concurrent deliveries to a destination, reuse collapsed to roughly one new connection per delivery.

## What changed

**One client per provider.** Nothing in the client configuration varies per destination — user agent, proxy settings and the transport wrapper all come from provider-level options fixed at registration — so the client moves from `CreatePublisher` to the provider constructor (`destwebhook`, `destwebhookstandard`, `desthookdeck`). This also means publisher-cache eviction no longer throws away the connection pool.

The rule the code now documents: only connection-level concerns justify a separate client. If per-destination proxy or client certificates are needed later, transports should be keyed by configuration and shared within a key.

**Derived pool sizing** (`internal/destregistry/connpool.go`) — sane defaults scaling with `DELIVERY_MAX_CONCURRENCY` (call it C):

| | value |
|---|---|
| per-host (depth) | `max(C, 2)` — 2 is Go's default, so C=1 never sizes below stock |
| total (breadth) | `clamp(32 × C, 512, max(4096, C))` |

`MaxIdleConns` caps only parked (idle) connections, never active ones — exceeding it causes connection churn, not errors. That makes the constants safe to reason about as reuse-rate tuning:

- **32 × C** ≈ how many distinct destinations a worker revisits within the 90s `IdleConnTimeout` at ~3s per delivery — slow destinations are where reuse matters most.
- **Floor 512** covers low-concurrency fanout: concurrency bounds simultaneous requests, not distinct hosts touched over time. C=1 at ~100ms per delivery still cycles ~900 destinations per idle window.
- **Cap 4096** bounds the parked-FD/memory cost where the reuse hit rate decays — but the cap rises to C, so it never undersizes the pool below the concurrency level.

The `hookdeck` provider talks to one host, so it gets a depth-only pool.

**Why not derive from `RLIMIT_NOFILE`** (an earlier revision did): the limit is per-process and unreliable in containers — the `RLIM_INFINITY` quirk meant the most generous environments derived the *smallest* pools — and mq-provider publishers hold per-destination FDs the derivation never accounted for. Since the limit only governs idle connections, tying it to a hard resource ceiling was solving the wrong problem.

**No new configuration.** These are defaults that scale with the one knob operators already set for delivery throughput. If a workload needs different sizing, an explicit config knob can be exposed later — adding one is backward-compatible; removing one isn't. Meanwhile:

- resolved values logged at startup (`delivery_max_idle_conns`, `delivery_max_idle_conns_per_host`)
- new `outpost.delivery_connections` counter, dimensioned by `type` and `reused` — the signal that the ceiling is binding

## Tests

`internal/destregistry/httpclient_pool_test.go` counts TCP connections opened at an `httptest` server via `ConnState`, swept across concurrency levels against both fast and slow destinations — the two regimes fail differently (latency vs. ephemeral ports). Connections opened track the concurrency level, not the request count.

A control case runs the identical workload through a stock-default client so the assertion can't pass trivially: at concurrency 32, stock opens ~120 connections for 320 requests, sized opens 32.

`destwebhook_connpool_test.go` covers the provider half — 8 publishers, 40 requests, still bounded by concurrency.

## Behavior changes worth flagging

- An invalid `DESTINATIONS_WEBHOOK_PROXY_URL` now fails at startup instead of on first publish.
- Idle connections are held on the receiver for up to 90 seconds after a delivery. Those connections were opened anyway; pooling changes whether they're held afterward.
- The pool is shared across tenants. Connections are per-host so there's no correctness concern, but a busy tenant can evict a quiet tenant's idle connections; the per-host limit bounds the blast radius.

## Verification

`go test -short ./...` clean. Pool tests run at `-count=15` for flake.


alexluong:
~~Added #1022 in d4d1175~~ — pulled back out. #1025 rewrites the same signature formatter constructors this touched, so keeping the two together meant one PR blocking the other on a merge conflict for no reason. #1022 will land separately, based on #1025.

This PR is #1017 only: connection pooling.

alexbouchardd (approved):

Issue #1014, opened 2026-08-03
retrymq: reduce Redis load from idle retry scheduler polling


alexluong:
Follow-up to #486, which added `RETRY_POLL_BACKOFF_MS` as a quick fix. This is the structural fix discussed there ("if there are no retries, we progressively back off").

## Current behavior

`scheduler.Monitor` calls `rsmq.ReceiveMessage` in a loop, sleeping `pollBackoff` (default 100ms) whenever no message is returned. Every RSMQ method first calls `getQueue` (`rsmq.go:299`), a `MULTI/HMGET/TIME/EXEC` pipeline that fetches `vt`/`delay`/`maxsize` and the server clock, purely to compute the arguments for the `EVALSHA` that follows.

One idle poll:

```
RTT 1:  MULTI / HMGET {ns}:deliverymq-retry:Q vt delay maxsize / TIME / EXEC
RTT 2:  EVALSHA <receiveMessage> → ZRANGEBYSCORE finds nothing → {}
```

5 client-observable commands, 2 round trips, whether or not anything is queued.

| | per second | per 30 days |
|---|---|---|
| Poll iterations | 10 | 25.9M |
| Redis commands | 50 | 129.6M |
| Round trips | 20 | 51.8M |

Counted as commands sent by the client. #486 cites 7/call, which counts `redis.call()`s inside the Lua body — the basis Upstash bills on. Both bases move by the same factor under this proposal.

This cost is per monitor instance and independent of traffic. Every replica runs its own monitor loop, so total Redis load scales with the number of running instances rather than with retry volume, and a queue that is empty for days costs exactly as much as a busy one.

`vt`/`delay`/`maxsize` are written once at `CreateQueue` and never mutated — outpost never calls `SetQueueAttributes` — so RTT 1 fetches three constants and a clock reading.

## Why upstream does it this way

RSMQ (Node, 2013) predates `redis.replicate_commands()` (Redis 3.2) and effect-replication-by-default (Redis 5). Under verbatim script replication, calling `TIME` inside a script would desync replicas, so passing the timestamp in from the client was the only option. That constraint is long gone; the split round trip survived the port to `go-rsmq` and into this fork.

## Proposal

**Part A — make `receiveMessage` self-contained.** Read `vt` from the `:Q` hash and call `TIME` inside the script. Eliminates RTT 1. 5 commands → 1, 2 round trips → 1.

[...]

PR #1026, opened 2026-08-07
perf(retrymq): cut idle Redis polling to one command per interval

alexluong:
Implements #1014. Makes the rsmq `receiveMessage` script self-contained — it reads `vt` from the `:Q` hash and calls `TIME` itself, dropping the `MULTI/HMGET/TIME/EXEC` that preceded every receive (5 commands / 2 RTTs → 1 / 1). When nothing is due the script also returns the time until the zset's earliest score, so the monitor sleeps until the next message comes due instead of every 100ms. At the new default that's 129.6M → 86.4K commands per month, per monitor.

`RETRY_POLL_BACKOFF_MS` keeps its meaning — how long the monitor waits when idle — and is redefined as the *maximum* idle sleep, default `100` → `0`, a sentinel meaning auto. Auto resolves to `min(30s, shortest configured retry delay)` (`GetRetryPollBackoff`), which makes added retry latency zero by construction for every config, including a custom `retry_schedule` with sub-30s entries. An explicit positive value is honored as-is — no cap — so a user can knowingly trade retry latency for idle cost (an earlier revision clamped explicit values to the shortest retry delay, which could silently *lower* a deliberately large value and raise the user's Redis cost). For any fixed value the new sleep is `min(nextDue − now, X)` ≤ the old flat `X`, so existing values only improve. Startup validation now rejects `retry_schedule` entries < 1, `retry_interval_seconds` < 1 (when no schedule is set), and negative `retry_poll_backoff_ms`.

**Heads-up — the consecutive-error ladder no longer derives from `pollBackoff`:** it now uses an internal 100ms base, so the ~1 minute of transient-infra tolerance documented at `scheduler.go` holds however the idle interval is configured. Previously raising the poll interval silently stretched (and lowering it shrank) that window.

## Behavior changes (release notes)

- **`RETRY_POLL_BACKOFF_MS` default `100` → `0` (auto).** Auto sleeps until the next due message, at most `min(30s, shortest configured retry delay)`. Retry timing is unchanged by construction; the idle Redis command rate drops ~600× (10 polls/s × 5 commands → ~1 command per interval).
- **Explicit values are now a fixed maximum idle sleep, honored as-is.** No clamping. A value larger than your shortest retry delay can add up to that much latency to retries scheduled while the monitor sleeps — that's now an explicit tradeoff you opt into, not a value we silently rewrite.
- **Stricter config validation.** Configs with `retry_schedule` entries < 1, `retry_interval_seconds` < 1 (when no schedule is set), or a negative `retry_poll_backoff_ms` now fail at startup instead of being silently accepted.
- **Crash-redelivery edge.** A message re-becoming visible after a crashed consumer (visibility timeout expiry) can take up to the idle sleep to be noticed, vs ~100ms before. Normal retry delivery is unaffected.

```sh
go test ./internal/rsmq/ ./internal/scheduler/ ./internal/config/   # Dragonfly
TESTCOMPAT=1 go test ./internal/rsmq/                               # + real Redis
```

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -81,7 +81,7 @@ type Config struct {
 	RetrySchedule                 []int `yaml:"retry_schedule" env:"RETRY_SCHEDULE" envSeparator:"," desc:"Comma-separated list of retry delays in seconds. If provided, overrides retry_interval_seconds and retry_max_limit. Schedule length defines the max number of retries. Example: '5,60,600,3600,7200' for 5 retries at 5s, 1m, 10m, 1h, 2h." required:"N"`
 	RetryIntervalSeconds          int   `yaml:"retry_interval_seconds" env:"RETRY_INTERVAL_SECONDS" desc:"Interval in seconds for exponential backoff retry strategy (base 2). Ignored if retry_schedule is provided." required:"N"`
 	RetryMaxLimit                 int   `yaml:"retry_max_limit" env:"MAX_RETRY_LIMIT" desc:"Maximum number of retry attempts for a single event delivery before giving up. Ignored if retry_schedule is provided." required:"N"`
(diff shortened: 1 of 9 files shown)


alexbouchardd:
See comment on issue regarding next message timestamp

alexbouchardd (approved):

PR #1033, opened 2026-08-11
Support custom dead-letter-queue names for self-provisioned SQS/RabbitMQ queues

ambroziepaval:
> Note: I haven't opened an issue for this first; happy to file one if that's preferred, or to close/rework this MR if you would rather take a different approach. Posting the implementation directly since it was small enough to just try out. Open to feedback, alternative designs if this direction makes sense.

**Why**
When Outpost self-provisions its internal delivery and log queues, it also auto-creates a DLQ for each, but the DLQ name is currently hardcoded as a derived suffix (`<queue>-dlq` for SQS, `<queue>.dlq` for RabbitMQ), with no way to override it.

If you're self-provisioning, it seems reasonable that you'd want to control the name of every resource Outpost creates, not just the main queues. Our organization has its own naming standard for dead-letter queues driven by infra/security policy (`dead_letter-<queue_name>`), which doesn't match what's assumed here, and we suspect other self-hosters run into the same kind of mismatch with their own conventions. Without an override, the only options are to fight the convention or build workarounds outside Outpost just to get provisioning to comply.

**Wanted to check**: is this something you'd be open to supporting, and if so, does this approach look right, or would you prefer a different shape (e.g. a full name-pattern/template instead of a fixed override, or a single shared suffix config instead of per-queue)?

**What changed**
Added optional config fields, symmetric to the existing `*_DELIVERY_QUEUE` / `*_LOG_QUEUE` settings, that let operators name each DLQ explicitly:
- `AWS_SQS_DELIVERY_DLQ` / `AWS_SQS_LOG_DLQ`
- `RABBITMQ_DELIVERY_DLQ` / `RABBITMQ_LOG_DLQ`

When unset, behavior is unchanged — Outpost falls back to the existing derived name (`<queue>-dlq` / `<queue>.dlq`), so this is purely additive and backward compatible.

### DIFF
--- a/internal/config/mqconfig_aws.go
+++ b/internal/config/mqconfig_aws.go
@@ -20,6 +20,8 @@ type AWSSQSConfig struct {
 	Endpoint        string `yaml:"endpoint" env:"AWS_SQS_ENDPOINT" desc:"Custom AWS SQS endpoint URL. Optional, typically used for local testing (e.g., LocalStack)." required:"N"`
 	DeliveryQueue   string `yaml:"delivery_queue" env:"AWS_SQS_DELIVERY_QUEUE" desc:"Name of the SQS queue for delivery events." required:"N"`
 	LogQueue        string `yaml:"log_queue" env:"AWS_SQS_LOG_QUEUE" desc:"Name of the SQS queue for log events." required:"N"`
+	DeliveryDLQ     string `yaml:"delivery_dlq" env:"AWS_SQS_DELIVERY_DLQ" desc:"Name of the dead-letter queue for the delivery queue. Optional; defaults to '<delivery_queue>-dlq' if unset." required:"N"`
+	LogDLQ          string `yaml:"log_dlq" env:"AWS_SQS_LOG_DLQ" desc:"Name of the dead-letter queue for the log queue. Optional; defaults to '<log_queue>-dlq' if unset." required:"N"`
 }
 
 func (c *AWSSQSConfig) getQueueName(queueType string) string {
@@ -28,6 +30,17 @@ func (c *AWSSQSConfig) getQueueName(queueType string) string {
 		return c.DeliveryQueue
 	case "logmq":
 		return c.LogQueue
+	default:
+		return ""
+	}
+}
+
+func (c *AWSSQSConfig) getDLQName(queueType string) string {
+	switch queueType {
+	case "deliverymq":
+		return c.DeliveryDLQ
+	case "logmq":
+		return c.LogDLQ
 	default:
 		return ""
 	}
@@ -44,6 +57,7 @@ func (c *AWSSQSConfig) ToInfraConfig(queueType string) *mqinfra.MQInfraConfig {
 			Region:                    c.Region,
 			ServiceAccountCredentials: c.getCredentials(),
 			Topic:                     c.getQueueName(queueType),
+			DLQ:                       c.getDLQName(queueType),
 		},
 	}
 }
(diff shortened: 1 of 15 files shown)


alexluong:
hi there, thanks for opening the PR, I think the motivation is reasonable. Let me think through the approach and share more soon!

internal/config/mqconfig_aws.go
```diff
@@ -20,6 +20,8 @@ type AWSSQSConfig struct {
 	Endpoint        string `yaml:"endpoint" env:"AWS_SQS_ENDPOINT" desc:"Custom AWS SQS endpoint URL. Optional, typically used for local testing (e.g., LocalStack)." required:"N"`
 	DeliveryQueue   string `yaml:"delivery_queue" env:"AWS_SQS_DELIVERY_QUEUE" desc:"Name of the SQS queue for delivery events." required:"N"`
 	LogQueue        string `yaml:"log_queue" env:"AWS_SQS_LOG_QUEUE" desc:"Name of the SQS queue for log events." required:"N"`
+	DeliveryDLQ     string `yaml:"delivery_dlq" env:"AWS_SQS_DELIVERY_DLQ" desc:"Name of the dead-letter queue for the delivery queue. Optional; defaults to '<delivery_queue>-dlq' if unset." required:"N"`
+	LogDLQ          string `yaml:"log_dlq" env:"AWS_SQS_LOG_DLQ" desc:"Name of the dead-letter queue for the log queue. Optional; defaults to '<log_queue>-dlq' if unset." required:"N"`
```
alexluong, line 24: we should add these to the startup log line in `internal/config/logging.go` (same for RabbitMQ and other cases)

internal/mqinfra/awssqs.go
```diff
@@ -15,6 +15,13 @@ type infraAWSSQS struct {
 	cfg *MQInfraConfig
 }
 
+func (infra *infraAWSSQS) dlqName() string {
```
alexluong, line 22: I think we should move this logic to `getDLQName()` instead? This would make the name available for the log line (commented above)

```golang
func (c *AWSSQSConfig) getDLQName(queueType string) string {
	switch queueType {
	case "deliverymq":
		return c.DeliveryDLQ
	case "logmq":
		return c.LogDLQ
	default:
		return ""
	}
}
```

alexluong (commented):
This approach works great! Can we apply the same approach to GCP and Azure as well? For Azure, I believe DLQ is a part of the subscription itself. I think we should add some sort of comment or something to make it explicit that we don't accept DLQ config instead of forgetting about it.

If you're not comfortable with GCP/Azure, I can follow up on that for you as well, no worries. Let me know how you'd like to proceed.

internal/mqinfra/awssqs.go
ambroziepaval: resolved in `getDLQName()` as suggested, with one deviation: dropping the mqinfra fallback made the DLQ field effectively required, which breaks direct constructors. So each provider now exports one definition of its convention (`DefaultAWSSQSDLQName` / `DefaultRabbitMQDLQName`), used by both `getDLQName()` and the existing fallback: field stays optional, suffix defined once.

ambroziepaval:
Azure: confirmed there's no DLQ entity to name: Declare only sets `DeadLetteringOnMessageExpiration` + `MaxDeliveryCount`, and consumption uses `azservicebus.SubQueueDeadLetter` (Azure's fixed $DeadLetterQueue)

GCP: want your call on the shape first, since it creates two DLQ resources and already names topics/subscriptions independently:

1. `*_DLQ_TOPIC` + `*_DLQ_SUBSCRIPTION` (4 vars) — consistent within GCP
2. `*_DLQ` only, subscription stays derived as <dlq>-sub — parity with SQS/RabbitMQ

alexluong:
Let's go with option 1!

> *_DLQ_TOPIC + *_DLQ_SUBSCRIPTION (4 vars) — consistent within GCP

alexluong (approved):
👌 🙏

PR #1036, opened 2026-08-11
chore: upgrade docker base image to debian13

8BitJonny:
debian12 has reached security end of life see https://endoflife.date/debian

so this is a simple PR to update the docker files to debian13


alexluong (approved):
thank you for the PR, LGTM!

PR #1038, opened 2026-08-13
feat(cli): add outpost config list command

### LINKED ISSUE #188: Expose Outpost config via CLI or API
As Outpost is quite configuration-driven, I think it could be useful to provide some mechanism for operators to figure out the config the service is running, either via a CLI command, API endpoint, or both. Since we do some parsing to get the final config value, it could come from the default value, a YAML config file, or environment variables, so a mechanism like this may provide some clarity.

Some CLI ideas:

```
$ outpost config list
$ outpost config get some.yaml.value
$ outpost config get SOME_ENV_VALUE
```

zahradm:
Adds `outpost config list` to print the effective, resolved config with secrets masked — useful for debugging why the server won't start, since it works  even when required fields are missing.
Closes #188

### DIFF
--- a/cmd/outpost/config.go
+++ b/cmd/outpost/config.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/hookdeck/outpost/internal/config"
+	"github.com/urfave/cli/v3"
+)
+
+// newConfigCommand builds the `outpost config` subcommand tree.
+func newConfigCommand() *cli.Command {
+	return &cli.Command{
+		Name:  "config",
+		Usage: "Inspect the effective configuration",
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:    "config",
+				Aliases: []string{"c"},
+				Usage:   "Path to config file",
+				Sources: cli.EnvVars("CONFIG"),
+			},
+		},
+		Commands: []*cli.Command{
+			{
+				Name:   "list",
+				Usage:  "Print the effective, resolved configuration with secrets masked",
+				Action: runConfigList,
+			},
+		},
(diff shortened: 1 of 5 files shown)


cmd/outpost/config.go
```diff
@@ -0,0 +1,57 @@
...
+	// often trying to see why the server won't start (a missing required
+	// field), so failing validation must not also hide the resolved config.
+	cfg, err := config.LoadWithoutValidation(flags)
+	if err != nil {
+		return fmt.Errorf("load config: %w", err)
+	}
+
+	// Reuse the same masked field set written to the startup logs, so the
+	// CLI output and the logs can never drift apart.
+	printConfigList(os.Stdout, cfg.LogConfigurationSummary())
```
alexluong, line 51: My main concern here is the `cfg.LogConfigurationSummary()` is not an exhaustive list of ALL config. Not sure if there's a more straightforward approach tho.

Ideally, we can log the cfg as is, but that requires some `SensitiveString` approach (#480). We're not yet ready for that yet.

I think summary is a good start, just want to highlight that aspect.

alexluong (commented):
Thank you for opening the PR, we appreciate the contribution!

Can you share more if this is an issue you're running into when working with Outpost? Just curious about the general context so we understand the needs better. Thanks!

zahradm:
Thank you for your response @alexluong 
This is my first contribution. I picked this up from the open issue. But the reason this feature is useful: if the server won't start because some setting is missing, right now there's no easy way to see what settings it actually loaded. This command shows you that.

alexluong:
I see. Glad to have you working on Outpost. In terms of the PR, I think it works great for now and we can proceed accordingly. I'll merge and include this in the next release shortly. Thanks again!

alexluong (approved):

PR #1051, opened 2026-08-26
fix(topics): honor wildcard flag during matching

### LINKED ISSUE #991: TOPICS_ALLOW_WILDCARDS: disabling the flag leaves existing wildcard subscriptions half-working
Follow-up to #912 (wildcard topic subscriptions, flag `TOPICS_ALLOW_WILDCARDS`).

The flag is only consulted at validation time (`Topics.Validate`, called from the destination create/update handlers). Topic matching on the delivery path (`Topics.MatchTopic` → `matchTopicPattern`) honors `*` unconditionally.

## Consequences of flipping enabled → disabled with existing wildcard destinations

1. **Existing wildcard subscriptions keep matching and delivering.** Disabling the flag stops new wildcards from being created but does not stop stored ones from expanding on the hot path.
2. **Those destinations become un-updatable.** Update re-validates the full destination, so any PATCH — including one that doesn't touch topics — fails with invalid topics until the caller also replaces the wildcard topics.
3. **Portal inconsistency.** The portal receives the flag and stops offering wildcard input, but existing wildcard subscriptions still render and behave as wildcards.

Net: the toggle reads like a kill switch but behaves as "stop new wildcards" with grandfathered data that delivers fine yet can't be edited.

## Possible mitigations to evaluate

- Only validate topics on update when the update actually changes them (removes the PATCH trap, keeps grandfathering) — likely the minimal fix.
- Gate `MatchTopic` on the flag so wildcards stop matching when disabled — makes the flag a true kill switch, but silently drops deliveries for existing subscriptions.
- Document the current semantics if grandfathering is the intended behavior.

Abdulmumin1:
## summary
​
makes `TOPICS_ALLOW_WILDCARDS` apply during topic matching, not only validation.
​
when the flag is disabled:
​
- persisted wildcard patterns like `user.*` are ignored during delivery matching
- exact topic subscriptions continue to work
- the standalone `*` still means subscribe to all topics
- stored wildcard topics are left untouched, so re-enabling the flag restores them
​
this also applies to destination filtering, direct destination publishes, and manual retries.
​
closes #991

### DIFF
--- a/docs/content/features/topics.mdoc
+++ b/docs/content/features/topics.mdoc
@@ -35,7 +35,7 @@ Wildcard topic subscriptions are disabled by default. Set `TOPICS_ALLOW_WILDCARD
 {% /tab %}
 {% /tabs %}
 
-When wildcard topic subscriptions are disabled, Outpost ignores stored wildcard patterns during matching. The stored topics are not changed, so re-enabling wildcard subscriptions restores the patterns. Exact topic subscriptions and the standalone `"*"` subscribe-to-all value continue to work.
+When wildcard topic subscriptions are disabled, Outpost ignores stored wildcard patterns during matching and omits them from destination and tenant API responses. The stored topics are not changed, so re-enabling wildcard subscriptions restores the patterns. Exact topic subscriptions and the standalone `"*"` subscribe-to-all value continue to work.
 
 ## Destination Topic Subscriptions
 
--- a/docs/content/features/topics.mdoc
+++ b/docs/content/features/topics.mdoc
@@ -35,6 +35,8 @@ Wildcard topic subscriptions are disabled by default. Set `TOPICS_ALLOW_WILDCARD
 {% /tab %}
 {% /tabs %}
 
+When wildcard topic subscriptions are disabled, Outpost ignores stored wildcard patterns during matching. The stored topics are not changed, so re-enabling wildcard subscriptions restores the patterns. Exact topic subscriptions and the standalone `"*"` subscribe-to-all value continue to work.
+
 ## Destination Topic Subscriptions
 
 Each destination subscribes to one or more topics. A destination receives an event only if:
--- a/docs/content/self-hosting/configuration.mdoc
+++ b/docs/content/self-hosting/configuration.mdoc
@@ -81,7 +81,7 @@ Choose one for event log persistence:
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `TOPICS` | — | Comma-separated list of topics your instance supports |
-| `TOPICS_ALLOW_WILDCARDS` | `false` | Allow `*` inside destination topic subscriptions, such as `user.*` |
+| `TOPICS_ALLOW_WILDCARDS` | `false` | Allow `*` inside destination topic subscriptions, such as `user.*`. When disabled, stored wildcard patterns are ignored but not deleted. |
 
 ## Portal
 
--- a/internal/apirouter/destination_displayer.go
+++ b/internal/apirouter/destination_displayer.go
@@ -1,20 +1,36 @@
 package apirouter
 
 import (
+	"strings"
+
 	"github.com/hookdeck/outpost/internal/destregistry"
 	"github.com/hookdeck/outpost/internal/models"
 )
 
 type destinationDisplayer struct {
-	registry destregistry.Registry
+	registry             destregistry.Registry
+	topicsAllowWildcards bool
 }
 
-func newDestinationDisplayer(r destregistry.Registry) *destinationDisplayer {
-	return &destinationDisplayer{registry: r}
+func newDestinationDisplayer(r destregistry.Registry, topicsAllowWildcards bool) *destinationDisplayer {
+	return &destinationDisplayer{
+		registry:             r,
+		topicsAllowWildcards: topicsAllowWildcards,
+	}
 }
 
 func (d *destinationDisplayer) Display(dest *models.Destination) (*destregistry.DestinationDisplay, error) {
-	return d.registry.DisplayDestination(dest)
+	display, err := d.registry.DisplayDestination(dest)
+	if err != nil {
+		return nil, err
+	}
+	if !d.topicsAllowWildcards {
+		displayDestination := *display.Destination
+		displayDestination.Topics = filterWildcardTopicPatterns(display.Destination.Topics)
+		display.Destination = &displayDestination
+	}
+
+	return display, nil
 }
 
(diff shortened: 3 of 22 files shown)


alexbouchardd:
That should also impact the destination and tenant object responses and in the postprocess trim out the no longer valid wildcard topics

Abdulmumin1:
> That should also impact the destination and tenant object responses and in the postprocess trim out the no longer valid wildcard topics

like before returning the object through the API, remove embedded wildcard topics such as user.* when the flag is disabled.

for example, storage contains:

["user.*", "order.created"]

with the flag disabled, the API return just:

["order.created"]

alexbouchardd:
Yes, otherwise the data is misrepresentative of the behavior and breaks the portal UI!

alexluong:
Hey @Abdulmumin1, sorry for the delayed review, and thanks for picking this up.

I tested this locally, flipping the flag on, off, and back on against the same data. It works as described, and re-enabling the flag restores everything.

> That should also impact the destination and tenant object responses and in the postprocess trim out the no longer valid wildcard topics

Picking up on this from @alexbouchardd, because I think it points at the underlying problem.

Outpost assumes every destination subscribes to at least one valid topic. Create and update enforce it (`topics: []` is rejected), and the API, the portal, and API clients all rely on it.

With this change, whether a topic is valid depends on the flag. When the flag is off, `user.*` is no longer a valid topic, so a destination that subscribes only to wildcards effectively subscribes to nothing. That breaks the assumption, and we can't reject the state the usual way because the data was valid when it was written.

The cases below are mostly that one violation showing up in different places. 1, 2 and 3 are direct results of a destination ending up with no effective topics. 4 comes from hiding topics the user can then overwrite. 5 is separate and minor.

Setup for all cases: `TOPICS=user.created,user.updated,user.deleted,order.created`, destinations created while the flag is `true`, observations made after switching it to `false`.

<details>
<summary><b>1. A wildcard-only destination looks healthy but receives nothing</b></summary>

**Repro**
1. Create a destination with `topics: ["user.*"]`.
2. Disable the flag.
3. Open it in the portal.

**Observed**
- List view: status `Active`, topics count `0`.
- Overview: the Topics row is blank.
- Settings: the topic picker has nothing selected.
- Publishing `user.created` does not reach it.

**Why it matters**
The end user sees an active destination with no topics and no explanation. They didn't change anything, so they have no way to tell why deliveries stopped. The operator has the same problem from the API: `topics: []` is indistinguishable from a destination that lost its data.

**Options**
- Docs only: state that wildcard-only destinations stop receiving events and show no topics.
- Portal hint on destinations with an empty topic list ("No active topic subscriptions").
- Expose the hidden state in the API response so clients can explain it, such as a boolean that says some topics are inactive.
</details>

<details>
<summary><b>2. GET returns a topics value that PATCH rejects</b></summary>

**Repro**
1. Create a destination with `topics: ["user.*"]`.
2. Disable the flag.
3. `GET /tenants/:id/destinations/:id` → `"topics": []`
4. `PATCH` the same destination with `"topics": []` → `422 validation failed: invalid topics`

**Observed**
The API returns a value for `topics` that it does not accept back.

**Why it matters**
Clients that read a destination and write the whole object back fail on these destinations: Terraform-style reconcilers, or an SDK update built from the read model. A PATCH that omits `topics` works fine, so this only affects full-object writers.

**Options**
- Leave as is and document it.
- Accept `topics: []` on update as "no change" when the flag is off and the stored topics are all wildcards.
- Related to case 4: whichever direction is picked there should probably apply here too.
</details>

<details>
<summary><b>3. An exact topic folded into a wildcard at write time is gone when the flag is off</b></summary>

**Repro**
1. With the flag on, create a destination with `topics: ["user.*", "user.created"]`.
2. `Normalize()` stores it as `["user.*"]`, since the wildcard covers the exact entry.
3. Disable the flag.
4. `GET` → `"topics": []`. Publishing `user.created` does not reach it.

**Observed**
The customer explicitly subscribed to `user.created`, but that entry no longer exists in storage, so there is nothing left to match once `user.*` is ignored.

**Why it matters**
Two destinations that the customer configured almost identically behave differently after the flip: `["user.created"]` keeps delivering, `["user.*", "user.created"]` stops. The second customer was more explicit and ends up worse off.

**Options**
- Docs only: note that exact topics covered by a wildcard are folded at write time and stop matching with the flag off.
- Stop folding exact entries into wildcards in `Normalize()` (keep only exact-duplicate removal). Larger change, separate PR, and it doesn't help data that is already folded.
</details>

<details>
<summary><b>4. Editing topics while the flag is off permanently drops the hidden wildcards</b></summary>

**Repro**
1. Create a destination with `topics: ["user.*", "order.created"]`.
2. Disable the flag. `GET` → `["order.created"]`.
3. `PATCH` with `"topics": ["order.created"]` (the same value the API returned) → `200`.
4. Stored topics are now `order.created` only.
5. Re-enable the flag. `GET` → `["order.created"]`. `user.*` is not restored.

**Observed**
Any topics write while the flag is off replaces the stored list, including the entries the API was hiding.

**Why it matters**
The PR's promise is that disabling is reversible. That holds until someone saves topics. In the portal this only happens when the user changes the topic selection and saves (config, credentials, and enable/disable never send topics), so it takes a deliberate edit. Still, the user can't see what they are overwriting.

**Options**
- Docs only: "editing a destination's topics while wildcards are disabled replaces the stored list, including hidden wildcard entries."
- Preserve hidden wildcard entries on update when the flag is off (merge the incoming list with the stored wildcard entries). Keeps reversibility intact but makes a topics write non-obvious.
</details>

<details>
<summary><b>5. Retries scheduled before the flip still deliver (not tested, from reading the code)</b></summary>

**Scenario**
1. With the flag on, an event matches a `user.*` destination and the delivery fails, so a retry is scheduled.
2. Disable the flag.
3. The scheduled retry fires.

**Expected from the code**
Matching happens at publish time and in the manual retry handler. The delivery worker doesn't re-check topics, so the scheduled retry goes out. Manual retry of the same event returns `400 destination does not match event`, so the two retry paths disagree for that window.

**Why it matters**
Minor. It's bounded by the retry schedule and arguably correct (the event was legitimately matched when it was published). Mentioning it so the behaviour is a decision and not a surprise.

**Options**
- Leave as is.
- Document that in-flight retries complete after the flag is disabled.
</details>

I think the thing to settle first is how we want to treat a destination whose effective topic list is empty, since 1, 2 and 3 follow from that answer. Once that's decided, 4 and 5 can likely be covered with a few sentences in `docs/content/features/topics.mdoc`.

alexbouchardd:
IMO, that's enough of an edge case that I would leave as is. The user has some responsibility here in turning it off.

alexluong (approved):
Thanks @Abdulmumin1 for the PR!

PR #1053, opened 2026-08-31
Add NATS JetStream as a message queue provider

### LINKED ISSUE #1052: Add NATS JetStream as a message queue provider
The selection of a provider happens by checking each one’s IsConfigured() sequentially:

if c.AWSSQS.IsConfigured() {
	c.adapter = &c.AWSSQS
} else if c.AzureServiceBus.IsConfigured() {
	c.adapter = &c.AzureServiceBus
} else if c.GCPPubSub.IsConfigured() {
	c.adapter = &c.GCPPubSub
} else if c.RabbitMQ.IsConfigured() {
	c.adapter = &c.RabbitMQ
}

Four providers, all of which require separate brokers/services to run alongside Outpost.

For teams running Kubernetes-native infrastructure – which Outpost targets directly (deployments/kubernetes/, examples/kubernetes/) – NATS JetStream is usually already in the stack, or a natural fit for one: a single Go binary with built-in persistence, and without a need to maintain a separate broker cluster.
There is nothing wrong with the four available providers; this is an addition for a specific category of
infrastructure deployments.

**Proposed change**

An addition of a NATSConfig struct in the same MQConfigAdapter form as all
providers currently implement:

type NATSConfig struct {
	ServerURL       string `yaml:"server_url" env:"NATS_SERVER_URL" required:"C"`
	Stream          string `yaml:"stream" env:"NATS_STREAM"`
	DeliverySubject string `yaml:"delivery_subject" env:"NATS_DELIVERY_SUBJECT"`
	LogSubject      string `yaml:"log_subject" env:"NATS_LOG_SUBJECT"`
	DeliveryDLQ     string `yaml:"delivery_dlq" env:"NATS_DELIVERY_DLQ"`
	LogDLQ          string `yaml:"log_dlq" env:"NATS_LOG_DLQ"`
}

func (c *NATSConfig) IsConfigured() bool { return c.ServerURL != "" }

and wired into the same chain, with one more else if and the implementation
of internal/mqinfra (stream/consumer provisioning – JetStream does not have
a broker-native DLQ, so exhausted messages get explicitly republished to a dedicated DLQ
subject/stream) and internal/mqs (the Queue interface).

Absolutely no modifications of any of the current providers – simply another
branch among all the others, the same as RabbitMQ or SQS already exist independently of each other today.

It’s all done, with the unit tests to match RabbitMQ/SQS test coverage and .env.example/contributing/mq.md updated
according to the MQ implementation checklist. It’s working in production against a live multi-tenant
webhook gateway right now. Can submit a PR for this issue if the idea looks reasonable.

rohanshirsath-droadmap:
Closes #1052

## Summary
Adds NATS JetStream as a 5th internal MQ option, alongside RabbitMQ/SQS/GCP
PubSub/Azure Service Bus — following the exact same MQConfigAdapter/MQInfra
interface pattern the existing providers use. Purely additive: no existing
provider's code is touched.

## Why NATS
JetStream is a single lightweight Go binary with built-in persistence — no
separate broker cluster to operate. Good fit for teams already running lean,
Kubernetes-native infrastructure who don't want to stand up RabbitMQ just
for this.

## What's included (per the MQ contribution checklist)
- [x] `internal/mqs` interface implementation (`queue_nats.go`)
- [x] `internal/mqinfra` interface implementation (`nats.go`)
- [x] Unit tests for config DLQ-name derivation, queue defaults, and
      connect-failure handling
- [x] `.env.example` entries
- [x] `contributing/mq.md` updated (Supported MQs list + a NATS behavior
      section matching the AWS SQS/RabbitMQ ones)

## Testing
Running in production against a real multi-tenant webhook gateway,
verified end to end (publish → durable delivery → ack). All existing
tests pass unmodified (`go test -short ./...`).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -80,6 +80,14 @@
 # AZURE_SERVICEBUS_LOG_SUBSCRIPTION="outpost-log-sub"
 # AZURE_SERVICEBUS_CONNECTION_STRING="Endpoint=sb://azuresb;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;"
 
+## NATS JetStream
+# NATS_SERVER_URL="nats://nats:4222"
+# NATS_STREAM="outpost"
+# NATS_DELIVERY_SUBJECT="outpost.delivery"
+# NATS_LOG_SUBJECT="outpost.log"
+# NATS_DELIVERY_DLQ="dlq.outpost-delivery"
+# NATS_LOG_DLQ="dlq.outpost-log"
+
 # ============================== PublishMQ ==============================
 
 ## RabbitMQ
--- a/contributing/mq.md
+++ b/contributing/mq.md
@@ -25,6 +25,7 @@ This document will focus on the first two types of integrations. The third type,
 - [x] GCP PubSub
 - [x] Azure ServiceBus
 - [x] RabbitMQ (AMQP 0.9.1)
+- [x] NATS JetStream
 
 publishmq only:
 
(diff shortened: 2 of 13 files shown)


rohanshirsath-droadmap:
@alexluong @alexbouchardd would appreciate a look when you have a moment — this closes #1052 and follows the existing MQ provider pattern closely, happy to adjust based on any feedback.

alexbouchardd:
Thanks! Agreed that NATS is a good natural fit. https://github.com/hookdeck/outpost/pull/910 already adds support, but multi-tenancy has caused some concerns. Is there a specific reason for a new PR?

We'll review and merge whichever seems most appropriate!

Just to set expectations, we have a big backlog of work; it might take some time.

rohanshirsath-droadmap:
@alexbouchardd  Thank you for your fast response and bringing up #910!

Reviewed it carefully — I believe it solves two distinct things and not the
same slot. 
#910 is a PublishMQ implementation (reading from a user's NATS environment), and the problem statement there itself has listed "Internal MQ support (delivery/log)" as "Out of scope". And this PR does solve just that
— provides NATS as an internal MQ provider for Outpost's delivery/log queues, just like RabbitMQ/SQS/GCP/Azure already do.

That also explains why the multi-tenancy issue need not be applied here — the complexity of #910 is about setting up a separate NATS connection per customer tenant (account-specific credentials, watched accounts directory) to aggregate as a publish source. 

According to Alex's comment in #910, the design principle that needs to be followed is "Outpost consumes from one
publish source over one connection" — and per-tenant/per-account connections will increase with the tenants * replicas count. This PR is exactly this "single connection" design pattern but for a different integration point: one NATS connection per Outpost deployment, similar to how RabbitMQ is consumed currently.

To add more, even putting aside the differences between #910 and publish-mq, the multi-tenancy approach of this PR is not new to — it is already what RabbitMQ uses currently (one shared exchange/queue for all Outpost tenants, and isolation handled by destination routing layer of Outpost, not by the message broker).

No urgency on the backlog at all!

alexbouchardd:
Valid! We'll get back to this PR as soon as soon as we are done with some other big pieces of work

PR #1056, opened 2026-09-01
feat: optional pprof endpoints via pprof_enabled

alexluong:
Adds `pprof_enabled` / `PPROF_ENABLED` (default off). When set, `net/http/pprof` is mounted under `/debug/pprof/` on the service HTTP port, for every service type — so a delivery or log worker can be profiled in place, not just the API.

The handlers are unauthenticated, so this is strictly opt-in and meant for ports that aren't publicly reachable.

```sh
PPROF_ENABLED=true ./outpost   # or pprof_enabled: true in config
go tool pprof http://localhost:3333/debug/pprof/profile?seconds=30
go tool pprof http://localhost:3333/debug/pprof/heap
```

https://claude.ai/code/session_011ox7wgp77JimQcPkV3LDa1

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -55,6 +55,7 @@ type Config struct {
 	APIKey       string `yaml:"api_key" env:"API_KEY" desc:"API key for authenticating requests to the Outpost API." required:"Y"`
 	APIJWTSecret string `yaml:"api_jwt_secret" env:"API_JWT_SECRET" desc:"Secret key for signing and verifying JWTs if JWT authentication is used for the API." required:"Y"`
 	GinMode      string `yaml:"gin_mode" env:"GIN_MODE" desc:"Sets the Gin framework mode (e.g., 'debug', 'release', 'test'). See Gin documentation for details." required:"N"`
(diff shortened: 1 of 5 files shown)


alexbouchardd (commented):
Do that need to be documented seems oddly specific, who is that helpful for?

alexluong:
> Do that need to be documented seems oddly specific, who is that helpful for?

Sorry, I don't follow. Are you talking about the config documentation? We can certainly keep it shorter if that's the comment.

alexbouchardd:
I don't understand who this flag is for and in which context, which I imagine most people reading the config doc wouldn't either

alexluong:
I don't believe this flag is documented in the config doc. The PR doesn't edit any markdown file, and we don't generate config doc from code anymore.

As for what this flag is, it's to enabled Golang's [`pprof`](https://pkg.go.dev/net/http/pprof) which is a profiler. Every time I need to profile, I would have to add this snippet, build a new image, deploy, and then I can start profiling. With this PR, I can add a flag and it's good to go.

This is the 2nd or 3rd time that I've needed this so figured it would be a good addition.

Also it's a common-ish practice AFAIK, for example: https://github.com/elastic/apm-server/pull/8002

alexbouchardd (approved):

Issue #1027, opened 2026-08-10
Postgres: no retention or partition management for events/attempts, causing unbounded growth


miekassu:
## Summary

Outpost's Postgres `events` and `attempts` tables are declared as range-partitioned on `time`, but only DEFAULT partitions are created.

We could not find any code that creates, rotates, or drops time-based Postgres partitions, nor any built-in Postgres retention mechanism. This leaves operators implementing retention externally with row-by-row `DELETE`, with the associated dead-tuple, vacuum, and index-maintenance overhead.

This appears to be an incomplete part of the intended Postgres design rather than a new feature request.

## Evidence in the source

The Postgres schema already uses time-based partitioning:

```sql
PRIMARY KEY (time, id)
) PARTITION BY RANGE (time);

CREATE TABLE events_default PARTITION OF events DEFAULT;
```

The same pattern is used for `attempts`.

More importantly, `internal/logstore/pglogstore/README.md:367-373` explicitly lists:

```text
Easy data retention (drop old partitions)
```

as a benefit of the time-based partitioning design.

However, we could not find runtime or migration code that creates, attaches, detaches, or drops those partitions.

The retention documentation also appears unfinished: `docs/content/self-hosting/guides/event-delivery-log.mdoc:8` currently contains:

```text
Need to talk about data retention and retention policies.
```

## Request

It would be great to complete the Postgres retention design by adding:

1. **Automatic time-range partition management** — create future partitions and drop expired ones.
2. **A Postgres retention setting** implemented on top of those partitions.
3. **A supported migration path** for existing installations where data is already in the DEFAULT partitions.
4. **Documentation** describing the intended Postgres retention model.

## Questions

1. Is Postgres partition lifecycle management already planned or tracked?
2. Is there a recommended migration path for existing DEFAULT partitions?
3. Until native retention exists, what retention approach do you recommend for Postgres?

## Small unrelated finding

The primary-key index on `attempts_default` still appears to be named `deliveries_default_pkey`.

`000005_denormalize_attempts.up.sql:59-60` renames the table and partition but apparently not the constraint-backing index. Harmless, but slightly confusing when inspecting `pg_stat_user_indexes`.

alexluong:
Hi @miekassu, thanks for sharing a great writeup. I can confirm that everything you called out here is correct.

To _briefly_ answer your questions first:

1. Is Postgres partition lifecycle management already planned or tracked?
**Yes, we have #249 to track this. That said, it's not on the roadmap currently.**

[...]

PR #1058, opened 2026-09-05
docs: how to adopt partitioning on an existing PostgreSQL install

AlexeyShalaev:
Follows up on #1027, where @alexluong asked for this as a docs PR.

The retention section already suggests managing partitions yourself with something like pg_partman once PostgreSQL is running at volume. This adds what an operator needs to know before doing that on an existing install, and a worked path for two tools.

The two caveats come from Outpost's schema, not from any tool. Every row an existing install has written lives in `events_default` / `attempts_default`, and PostgreSQL won't attach a partition for a month whose rows are still in the DEFAULT partition, so adoption means moving rows first rather than only creating partitions going forward. And because the tables always carry a DEFAULT partition, `DETACH PARTITION ... CONCURRENTLY` is unavailable, so each retention drop takes a short `ACCESS EXCLUSIVE` on the table. Both are stated plainly so nobody plans a maintenance window around a lock-free drop that isn't.

Both recipes were run against the current `internal/migrator/migrations/postgres/*.up.sql` on PostgreSQL 17, with eight months of rows seeded into `events_default`. The pg_partman one took two attempts to get right, which is why it's worth having in the docs: `create_parent` creates its own `events_default`, so it collides unless the existing one is renamed first, and `partition_data_proc` refuses to run before the table is in `part_config`. The order in the snippet is the one that ends with the DEFAULT partition empty, 13 monthly partitions, and `run_maintenance` healthy. The pg-partsmith one uses the published image and Outpost's own `POSTGRES_URL`.

Scope: one page, no nav or redirect changes, one `{% callout %}` from the supported component list. Voice checked against `.agents/skills/hookdeck-outpost-docs-format`.

I maintain pg-partsmith, which is why I was reading this code in the first place. If you'd rather the page mention only pg_partman, I'm happy to trim it to that; the two caveats and the pg_partman ordering are the part I'd most want to survive.

### DIFF
--- a/docs/content/self-hosting/guides/event-delivery-log.mdoc
+++ b/docs/content/self-hosting/guides/event-delivery-log.mdoc
@@ -64,7 +64,58 @@ DELETE FROM events   WHERE time < now() - interval '30 days';
 
 Run these in bounded batches rather than as a single statement, on whatever scheduler you already operate, such as a Kubernetes `CronJob`, a systemd timer, or [pg_cron](https://github.com/citusdata/pg_cron) if the extension is available to you. Both tables have a primary key leading with `time`, so these are index range scans rather than full table scans. You'll still take the dead tuple and vacuum cost that any large delete incurs.
 
-The `events` and `attempts` tables are declared as range partitioned on `time`, but Outpost creates only DEFAULT partitions and doesn't manage the partition lifecycle. Dropping expired partitions is cheaper than deleting rows at scale, so if you're running PostgreSQL at volume, managing partitions yourself with a tool like [pg_partman](https://github.com/pgpartman/pg_partman) is worth considering.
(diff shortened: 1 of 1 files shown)


AlexeyShalaev:
@alexluong following up on this one — you asked for it in #1027, and I know a docs PR is an easy thing to slip past a week of feature work. It's still one file and still applies cleanly to main.

If the pg-partsmith half is what gives you pause, say so and I'll cut it. The two caveats and the pg_partman ordering are the part I'd want in the docs regardless, and they stand on their own without it.

alexluong:
Hi @AlexeyShalaev, thanks for the ping. Sorry I haven't got back. Confirming it's on my TODO and I'll aim to get back to you early next week.

PR #1060, opened 2026-09-09
feat: chain forward proxies in DESTINATIONS_WEBHOOK_PROXY_URL

alexluong:
`DESTINATIONS_WEBHOOK_PROXY_URL` accepts several whitespace-separated proxy URLs, nearest hop first. Outpost dials the first, sends one `CONNECT` per intermediate hop, and hands the tunneled connection to Go's transport with the last hop as its proxy, so the last hop keeps stdlib CONNECT, auth and TLS handling and a pooled connection is one open tunnel. Every hop only needs standard CONNECT support. A single URL parses and behaves exactly as before. Needed when our egress proxy must itself go through a destination's own allowlisting proxy.

Intermediate-hop failures follow the existing attribution rules: unreachable first hop or 401/403/407 anywhere is an infra nack, any other non-200 is a failed attempt attributed to the hop that could not be reached.

**Heads-up, 403 + `rbac_access_denied` (e9d30e25):** an Envoy RBAC deny is now a failed attempt with code `network_unreachable` on any hop, including the last. Previously every 403 was an infra nack, which would redeliver forever for a destination the egress policy blocks.

```sh
# .env: LOCAL_DEV_ENVOY=1 LOCAL_DEV_ENVOY_CHAIN=1
# DESTINATIONS_WEBHOOK_PROXY_URL="http://envoy:10000 http://envoy-upstream:10000"
make up && make smoke
docker compose -p outpost logs envoy | grep CONNECT            # -> envoy-upstream:10000
docker compose -p outpost logs envoy-upstream | grep CONNECT   # -> mock.hookdeck.com:443
```

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -26,6 +26,9 @@ LOCAL_DEV_RABBITMQ=1
 
 # Forward proxy for testing DESTINATIONS_WEBHOOK_PROXY_URL flows
 # LOCAL_DEV_ENVOY=1
+# Second proxy behind the first, for hop-list testing (needs LOCAL_DEV_ENVOY=1):
+#   DESTINATIONS_WEBHOOK_PROXY_URL="http://envoy:10000 http://envoy-upstream:10000"
+# LOCAL_DEV_ENVOY_CHAIN=1
 
 # Observability (pick one — they conflict on otel-collector + OTLP ports)
 # LOCAL_DEV_GRAFANA=1
--- a/.env.example
+++ b/.env.example
@@ -124,6 +124,8 @@
 # DELIVERYMQ_MAX_CONCURRENCY=1
 
 # Webhook Destination
+# Forward proxy for webhook deliveries.
+# DESTINATIONS_WEBHOOK_PROXY_URL="http://user:pass@proxy:10000"
 # DESTINATIONS_WEBHOOK_HEADER_PREFIX="x-outpost-"
 # Header name configs are three-state: unset uses the default <prefix>+key,
 # an explicit value pins that exact name, and an empty string disables the header.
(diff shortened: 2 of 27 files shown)


alexbouchardd:
Space-delimited instead of comma? 🤔

alexbouchardd (approved):

alexluong:
> Space-delimited instead of comma? 🤔

My thinking is that comma is technically valid in the proxy URL, for example if there's some sort of query params or even basic auth username/password.

On the other hand, an unencoded space can never appear in a valid URL.

alexbouchardd:
Make sense 👍

mkherlakian:
High level observation, maybe intentional - in chain where you set up proxy A -> proxy B -> proxy C

We treat connection failures (if I'm reading the code correctly) to A differently than B and onwards as we retry silently if A fails, but not if B,C etc fail, and I see it's also return code specific.

This works for Outpost, but is it how we want it to work for OSS? should we make it so that it's configurable in the deployment?

alexluong:
> High level observation, maybe intentional - in chain where you set up proxy A -> proxy B -> proxy C
> We treat connection failures (if I'm reading the code correctly) to A differently than B and onwards as we retry silently if A fails, but not if B,C etc fail, and I see it's also return code specific.

Yes, this is intentional, altho I do understand where you're coming from here.

> This works for Outpost, but is it how we want it to work for OSS? should we make it so that it's configurable in the deployment?

My take is that this proxy chain is a very edge case feature that doesn't necessarily warrant further configuration unless there's a clear use case. If I have to guess, most won't proxy, and even if they do it will have 1 hop only.

PR #1064, opened 2026-09-10
fix(deliverymq): Add index to attempts_default partition

samadalishah:
## Summary

Adds a PostgreSQL composite index on `attempts_default` to optimize retry execution lookups for the latest attempt by event, tenant, and destination.

## Context

Retry execution fetches the latest prior attempt from logstore before publishing the next delivery task.
`deliverymq-retry task → retry scheduler → fetch latest attempt → execute delivery`

The query filters by:

- `event_id`
- `tenant_id`
- `destination_id`

and orders by:

- `time DESC`
- `id DESC`

with `LIMIT 1`.

Under high I/O load, PostgreSQL can choose an existing tenant/destination/time index and apply `event_id` as a heap filter. In large default partitions, that can require many page reads for a single retry lookup.

During an incident, slow Postgres retry lookups contributed to retry tasks running longer than the Redis retry visibility timeout `RETRY_VISIBILITY_TIMEOUT_SECONDS` (default 30 secs). That allowed the same retry task to become visible again and increased duplicate retry pressure.

## Changes

Adds this index:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS attempts_default_event_tenant_destination_time_id_idx
ON attempts_default (event_id, tenant_id, destination_id, time DESC, id DESC);
```

##  Operational notes

This index is created only on `attempts_default`, which matches the current production table layout.
`CREATE INDEX CONCURRENTLY` is used to avoid blocking writes while the index is built. Building the index can still be I/O intensive on large deployments, so it should be monitored during rollout.
Longer term, monthly partitions would make retention and table maintenance safer, since old attempts could be removed by dropping partitions instead of batched deletes.

### DIFF
--- a/internal/migrator/migrations/postgres/000011_retry_attempt_lookup_index.down.sql
+++ b/internal/migrator/migrations/postgres/000011_retry_attempt_lookup_index.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX CONCURRENTLY IF EXISTS attempts_default_event_tenant_destination_time_id_idx;
--- a/internal/migrator/migrations/postgres/000011_retry_attempt_lookup_index.up.sql
+++ b/internal/migrator/migrations/postgres/000011_retry_attempt_lookup_index.up.sql
@@ -1,0 +1,5 @@
+-- Support retry scheduler lookups for the latest attempt by event, tenant, and
+-- destination. This is created on the default partition because PostgreSQL does
+-- not support CREATE INDEX CONCURRENTLY on a partitioned parent table.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS attempts_default_event_tenant_destination_time_id_idx
+ON attempts_default (event_id, tenant_id, destination_id, time DESC, id DESC);


samadalishah:
**For more context:**
We recently had an incident where a bunch of tasks went into `deliverymq-retry-dlq` and threw these logs:
`task exceeded max receive count, moved to dead-letter queue`

Our default partition already has 30 million rows. Looking into the **attempts** table, we noticed a particular event was attempted more than the allowed maximum (attempt_max: 6, but attempt_number: 11). This was happening for several events. 

The root cause was that the Postgres DB was overloaded due to storage I/O saturation. Because the lookup query took more than 30 seconds (and up to 40+ seconds in some instances) and the Redis `RETRY_VISIBILITY_TIMEOUT_SECONDS` default is 30 seconds, the same task became visible multiple times. This created a feedback loop, causing even more load on the DB.

As a quick fix, we scaled up the DB instance to alleviate the bottleneck. However, we found that adding this index would be a more permanent help.

We have already applied this index to our database, but we opened this PR to see what the general stance is for this case. Also, are there any plans to add data cleanup as a feature to Outpost?

alexluong:
Hi @samadalishah, thanks for opening the PR and sharing more context.

I'm trying to fully understand the issue. This statement strikes me as a bit odd.

> Under high I/O load, PostgreSQL can choose an existing tenant/destination/time index and apply event_id as a heap filter. In large default partitions, that can require many page reads for a single retry lookup.

I don't fully follow why this is the case. I'm not sure why this is the behavior, especially "under high I/O load" since I don't fully see how that's related.

I'm personally leaning more towards the query itself using `ANY` even tho there's only 1 item.

Do you have the plan of the query when it used the wrong index by chance?

samadalishah:
Hi @alexluong thank you for getting back on the PR so quickly 🙌

> I don't fully follow why this is the case. I'm not sure why this is the behavior, especially "under high I/O load" since I don't fully see how that's related.

Good callout! That sentence is worded wrong, my bad. The high I/O load was not meant to imply that Postgres changes its plan because the system is saturated. The issue is the query selection and existing index itself.

The retry lookup filters by `tenant_id`, `destination_id`, and `event_id`, then orders by `time DESC, id DESC` with `LIMIT 1`.

Today, Postgres can use the existing `idx_attempts_tenant_dest_time` index because it matches:
- `tenant_id`
- `destination_id`
- `ORDER BY time DESC, id DESC`

But that index does not include `event_id`, so `event_id` is applied as an additional filter while scanning candidate rows for that tenant/destination.

If a tenant/destination has many attempts, the query can scan a large number of recent rows before finding the matching `event_id`. During the incident, when the database was already I/O saturated, that access pattern became much more expensive and some retry lookups took 40+ seconds.

Unfortunately, I didn't keep the plan before applying the index but this is the current one

<details>
<summary><b>Current query plan</b></summary>

```json
[
  {
    "Plan": {
      "Node Type": "Limit",
      "Parallel Aware": false,
      "Async Capable": false,
      "Startup Cost": 8.72,
      "Total Cost": 8.72,
      "Plan Rows": 1,
      "Plan Width": 1000,
      "Actual Startup Time": 8.909,
      "Actual Total Time": 8.910,
      "Actual Rows": 1.00,
      "Actual Loops": 1,
      "Disabled": false,
      "Output": ["attempts.id", "attempts.event_id", "attempts.tenant_id", "attempts.destination_id", "attempts.destination_type", "attempts.topic", "attempts.status", "attempts.\"time\"", "attempts.attempt_number", "attempts.manual", "attempts.code", "attempts.response_data", "attempts.event_time", "attempts.eligible_for_retry", "attempts.event_data", "attempts.event_metadata"],
      "Shared Hit Blocks": 9,
      "Shared Read Blocks": 14,
      "Shared Dirtied Blocks": 0,
      "Shared Written Blocks": 0,
      "Local Hit Blocks": 0,
      "Local Read Blocks": 0,
      "Local Dirtied Blocks": 0,
      "Local Written Blocks": 0,
      "Temp Read Blocks": 0,
      "Temp Written Blocks": 0,
      "Shared I/O Read Time": 8.690,
      "Shared I/O Write Time": 0.000,
      "Local I/O Read Time": 0.000,
      "Local I/O Write Time": 0.000,
      "Temp I/O Read Time": 0.000,
      "Temp I/O Write Time": 0.000,
      "Plans": [
        {
          "Node Type": "Sort",
          "Parent Relationship": "Outer",
          "Parallel Aware": false,
          "Async Capable": false,
          "Startup Cost": 8.72,
          "Total Cost": 8.72,
          "Plan Rows": 1,
          "Plan Width": 1000,
          "Actual Startup Time": 8.907,
          "Actual Total Time": 8.908,
          "Actual Rows": 1.00,
          "Actual Loops": 1,
          "Disabled": false,
          "Output": ["attempts.id", "attempts.event_id", "attempts.tenant_id", "attempts.destination_id", "attempts.destination_type", "attempts.topic", "attempts.status", "attempts.\"time\"", "attempts.attempt_number", "attempts.manual", "attempts.code", "attempts.response_data", "attempts.event_time", "attempts.eligible_for_retry", "attempts.event_data", "attempts.event_metadata"],
          "Sort Key": ["attempts.\"time\" DESC", "attempts.id DESC"],
          "Sort Method": "top-N heapsort",
          "Sort Space Used": 26,
          "Sort Space Type": "Memory",
          "Shared Hit Blocks": 9,
          "Shared Read Blocks": 14,
          "Shared Dirtied Blocks": 0,
          "Shared Written Blocks": 0,
          "Local Hit Blocks": 0,
          "Local Read Blocks": 0,
          "Local Dirtied Blocks": 0,
          "Local Written Blocks": 0,
          "Temp Read Blocks": 0,
          "Temp Written Blocks": 0,
          "Shared I/O Read Time": 8.690,
          "Shared I/O Write Time": 0.000,
          "Local I/O Read Time": 0.000,
          "Local I/O Write Time": 0.000,
          "Temp I/O Read Time": 0.000,
          "Temp I/O Write Time": 0.000,
          "Plans": [
            {
              "Node Type": "Index Scan",
              "Parent Relationship": "Outer",
              "Parallel Aware": false,
              "Async Capable": false,
              "Scan Direction": "Forward",
              "Index Name": "attempts_default_event_tenant_destination_time_id_idx",
              "Relation Name": "attempts_default",
              "Schema": "public",
              "Alias": "attempts",
              "Startup Cost": 0.69,
              "Total Cost": 8.71,
              "Plan Rows": 1,
              "Plan Width": 1000,
              "Actual Startup Time": 1.486,
              "Actual Total Time": 8.854,
              "Actual Rows": 12.00,
              "Actual Loops": 1,
              "Disabled": false,
              "Output": ["attempts.id", "attempts.event_id", "attempts.tenant_id", "attempts.destination_id", "attempts.destination_type", "attempts.topic", "attempts.status", "attempts.\"time\"", "attempts.attempt_number", "attempts.manual", "attempts.code", "attempts.response_data", "attempts.event_time", "attempts.eligible_for_retry", "attempts.event_data", "attempts.event_metadata"],
              "Index Cond": "((attempts.event_id = ANY ('{f898320b-5aed-4ee8-b139-fdcc112f556d}'::text[])) AND (attempts.tenant_id = ANY ('{8b496aa5-6641-450f-8e9f-538c18615905}'::text[])) AND (attempts.destination_id = ANY ('{9e380e71-7f97-4e54-b44e-515279fc7760}'::text[])))",
              "Rows Removed by Index Recheck": 0,
              "Index Searches": 1,
              "Shared Hit Blocks": 3,
              "Shared Read Blocks": 14,
              "Shared Dirtied Blocks": 0,
              "Shared Written Blocks": 0,
              "Local Hit Blocks": 0,
              "Local Read Blocks": 0,
              "Local Dirtied Blocks": 0,
              "Local Written Blocks": 0,
              "Temp Read Blocks": 0,
              "Temp Written Blocks": 0,
              "Shared I/O Read Time": 8.690,
              "Shared I/O Write Time": 0.000,
              "Local I/O Read Time": 0.000,
              "Local I/O Write Time": 0.000,
              "Temp I/O Read Time": 0.000,
              "Temp I/O Write Time": 0.000
            }
          ]
        }
      ]
    },
    "Query Identifier": -5849889232272461497,
    "Planning": {
      "Shared Hit Blocks": 414,
      "Shared Read Blocks": 27,
      "Shared Dirtied Blocks": 0,
      "Shared Written Blocks": 0,
      "Local Hit Blocks": 0,
      "Local Read Blocks": 0,
      "Local Dirtied Blocks": 0,
      "Local Written Blocks": 0,
      "Temp Read Blocks": 0,
      "Temp Written Blocks": 0,
      "Shared I/O Read Time": 4.384,
      "Shared I/O Write Time": 0.000,
      "Local I/O Read Time": 0.000,
      "Local I/O Write Time": 0.000,
      "Temp I/O Read Time": 0.000,
      "Temp I/O Write Time": 0.000
    },
    "Planning Time": 6.130,
    "Triggers": [
    ],
    "Execution Time": 8.963
  }
]

```
</details>

alexluong:
Great, thanks for sharing @samadalishah.

My thinking here is to defer this PR for now given we don't have a strong pull on this. Hookdeck operates Outpost with ClickHouse, so we're more in-tune with the performance characteristics of that backend.

There are 2 reasons why we're a bit hesitant on this:

1. This index feels a bit unnecessary based on my understanding/analysis so far. It definitely does help but the root cause feels like a query problem rather than the indexing problem. The key here is that PG should have used the event_id index instead as that should be the main identifier.
2. We have plans for a different retry logic for performance reasons which will completely change this query in the next version. Therefore, I don't think this is suitable for us at this moment.

We can continue monitoring it if other users run into this issue. Since it's not a blocker as you can create the index manually, I think we can close this PR for now.

That said, if you'd like to dig deeper and provide more clarity on this, I'd be happy to continue the discussion further.

samadalishah:
Yes, that makes sense. We are also keeping eyes on the system and can come back again with more info if this happens. Also, since you mentioned that there are plans to revamp the retry feature, I am also leaning towards deferring this PR for now. Thanks again for the insights!

Issue #1068, opened 2026-09-17
RFC: Dual webhook signatures for migrating signature schemes


alexluong:
## Summary

Let a deployment send two webhook signature formats on the same request for a transition period. Operators can then move to Outpost from a webhook system they built themselves or from another provider, or change signature schemes, without breaking receivers' verification code.

## Motivation

A receiver's verification code is coupled to the signature format: header names, signed content, encoding, key derivation. Two cases change that format for existing receivers:

1. **Changing the scheme of a running deployment**, e.g. moving from a custom template to a different one, or renaming headers.
2. **Moving to Outpost from a previous webhook system**, self-built or another provider. Receivers verify that system's headers. Outpost sends its own.

Today both are hard cutovers. Every receiver has to deploy new verification code at the moment the operator flips the config.

## Approach

Add a second, optional set of webhook signature config. If it is set, Outpost signs each request twice and sends both sets of headers: the primary signature as today, plus the second one built from the alternative config. If it isn't set, nothing changes.

The second set takes the same options as the primary signature (header name, content template, header template, encoding, algorithm). When configured, it applies to every webhook destination in the deployment, until the operator removes the config.

## Naming

The second config set needs a name. Options:

- `DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_HEADER_NAME`
- `DESTINATIONS_WEBHOOK_LEGACY_SIGNATURE_HEADER_NAME`
- `DESTINATIONS_WEBHOOK_ALT_SIGNATURE_HEADER_NAME`
- `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_NAME_COMPAT` (suffix form)
- `DESTINATIONS_WEBHOOK_SIGNATURE_HEADER_NAME_1` (numbered)

The rest of this RFC uses `COMPAT` as a prefix group (`destinations.webhook.compat.*` in YAML). The mechanism doesn't depend on the term, and feedback on it is welcome.

## Proposal

The second signature set is called the **compat signature** below.

### Operator config

| Key | Description |
|---|---|
| `DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_HEADER_NAME` | Name of the compat signature header. Setting it enables the feature. |
| `DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_CONTENT_TEMPLATE` | Same semantics as `SIGNATURE_CONTENT_TEMPLATE` |
[...]

PR #1069, opened 2026-09-17
feat(destwebhook): compat signature for migrating signature schemes

alexluong:
Implements #1068: an optional second webhook signature, in the operator's previous format, sent on the same request as the primary one. Default webhook mode only; nothing changes unless `DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_HEADER_NAME` is set.

The compat set mirrors the primary signature options under `DESTINATIONS_WEBHOOK_COMPAT_*`, plus templated companion headers (`COMPAT_HEADERS`) and secret key decoding (`SIGNATURE_SECRET_ENCODING` / `_PREFIX`, also added to the primary signature). Per review, destination cutoff, per-destination setting and sunset timestamp are left out; the compat signature applies to every webhook destination until the config is removed.

Guide: `docs/content/guides/change-webhook-signature-scheme.mdoc`. e2e: `TestE2E_CompatSignature` verifies deliveries with the official Standard Webhooks Go SDK for the compat headers and the verify guide's algorithm for the primary ones.

```bash
# Standard Webhooks-format signature alongside the default x-outpost-signature
DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_HEADER_NAME=webhook-signature
DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_CONTENT_TEMPLATE='{{.EventID}}.{{.Timestamp.Unix}}.{{.Body}}'
DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_HEADER_TEMPLATE='v1,{{index .Signatures 0}}{{range slice .Signatures 1}} v1,{{.}}{{end}}'
DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_ENCODING=base64
DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_SECRET_ENCODING=base64
DESTINATIONS_WEBHOOK_COMPAT_SIGNATURE_SECRET_PREFIX=whsec_
DESTINATIONS_WEBHOOK_COMPAT_HEADERS='webhook-id={{.EventID}},webhook-timestamp={{.Timestamp.Unix}}'
```


docs/content/nav.json
```diff
@@ -86,6 +86,10 @@
             "slug": "guides/verify-webhook-signatures",
             "title": "Verify Webhook Signatures"
           },
+          {
+            "slug": "guides/migrate-webhook-signatures",
+            "title": "Migrate Webhook Signatures"
```
alexbouchardd: Worried this might be missintepreted into migrating signatures to Outpost which isn't the case. I don't think that's immediately obvious that this is a guide to change your signature format which is entirely optional

alexluong: reworded a bit, changed to "Change Webhook Signature Scheme". What do you think? Or any suggestion for this?

alexbouchardd (approved):

docs/content/nav.json
alexbouchardd: Looks good 👍

PR #1070, opened 2026-09-18
fix(portal): show empty destination metrics

alexbouchardd:
## Summary

- treat destinations omitted from a completed batched metrics response as an empty series
- render six empty histogram samples with a `0` total instead of leaving the loading skeleton visible

## Validation

- `npm run lint`
- `npm run build`
- rendered the portal against the reported response shape with two zero-delivery destinations and one destination with 34 deliveries

## Screenshot

Destination list showing six empty histogram samples for destinations with zero deliveries


internal/portal/src/scenes/DestinationsList/DestinationEventsCell.tsx
```diff
@@ -6,6 +6,10 @@ interface DestinationEventsCellProps {
   isLoading: boolean;
 }
 
+// The destination list queries 24 hours at 4-hour granularity. The API
+// includes both boundary buckets, so a complete sparkline has seven samples.
+const EMPTY_SERIES_SAMPLE_COUNT = 7;
```
alexluong: So super nit/edge case here but in theory, if the query is sent at a certain "boundary" minute like 00:00, 04:00, etc., the response will have 6 item instead of 7

Image

alexluong (approved):

internal/portal/src/scenes/DestinationsList/DestinationEventsCell.tsx
alexbouchardd: Simplified this further: the list always represents 24 hours in 4-hour slices, so an empty series now renders a fixed 6 samples. The request-derived bucket-count code has been removed.

alexluong: > the list always represents 24 hours in 4-hour slices

Not sure if I follow you correctly but I don't think this is correct.

The server truncates `time[start]` down to the bucket boundary, so a 24h window at 4h usually spans 7 buckets, not 6. Portal rounds `now` down to the minute before building the request, so any load between 04:00:00 and 04:00:59 sends `time[end]=04:00:00`.

**Request at 04:37 UTC → 7 buckets**

request: 09-18 04:37:00 → 09-19 04:37:00, start truncated to 09-18 04:00:00

1. 09-18 04:00 → 08:00 (partial, covers 04:37 onward)
2. 09-18 08:00 → 12:00
3. 09-18 12:00 → 16:00
4. 09-18 16:00 → 20:00
5. 09-18 20:00 → 09-19 00:00
6. 09-19 00:00 → 04:00
7. 09-19 04:00 → 08:00 (partial, covers up to 04:37)

**Request during 04:00:00–04:00:59 UTC → 6 buckets**

request: 09-18 04:00:00 → 09-19 04:00:00, start already aligned so truncation is a no-op

1. 09-18 04:00 → 08:00
2. 09-18 08:00 → 12:00
3. 09-18 12:00 → 16:00
4. 09-18 16:00 → 20:00
5. 09-18 20:00 → 09-19 00:00
6. 09-19 00:00 → 04:00

7 is the normal case. 6 only during the one minute after each 4h boundary, so 6 of every 1440 minutes.

So with a fixed 6, empty rows are one bar narrower than populated rows almost all the time. Original 7 was the better default. If we want it exact, take the sample count from any populated destination in the same response and fall back to 7 when all are empty.

alexbouchardd: You’re right. I removed the fixed-six behavior in follow-up #1073. Empty rows now take their sample count from the first populated destination in the same response, with 7 as the fallback when every destination is empty.

PR #1071, opened 2026-09-18
feat: Add config to enable Redis TLS verification

hogklint:
Seems like an oversight to not allow Redis TLS verification. With this commit it's just a matter of setting `REDIS_TLS_VERIFY=true`, and if a self-signed cert is used the environment variables `SSL_CERT_DIR` and `SSL_CERT_FILE` can be used.

```sh
# For example
SSL_CERT_DIR=/etc/ssl/certs:/custom-certs
# or
SSL_CERT_FILE=/custom-certs/redis.crt
```

Let me know if I've missed to include anything.


alexluong (approved):
Great, thank you for opening the PR. You're right. This is a good addition to Outpost. I've added a commit improving the docs on this as well, hope that's cool with you. Merging now!

hogklint:
> I've added a commit improving the docs on this as well, hope that's cool with you

For sure! Thanks for fixing that :+1:

PR #1079, opened 2026-09-22
feat(metrics): per-attempt delivery latency and latency measures on /metrics/attempts

kunwar-axilio:
Implements the proposal agreed in #1074: per-attempt delivery latency, persisted on the attempt and queryable through the attempt metrics API.

## Why

The metrics API exposes activity and error rate per destination, but no response time. The only latency Outpost emits today is the `outpost.delivery_latency` OTel histogram, which is end-to-end (publish → delivered), labelled by destination *type* only, and not queryable per destination or tenant. Anyone building a per-destination dashboard on `/metrics/attempts` can show volume and failures but not "how fast is this endpoint responding, and is its p95 drifting". This closes that gap without touching the OTel side (no new high-cardinality labels).

## What changed

Decisions follow the discussion thread: per-attempt HTTP response time (not end-to-end), nullable column, `null` rather than `0` in empty buckets, p50/p95/p99, captured at the registry for all providers.

**Capture** — `destregistry.PublishEvent` times the provider's `Publish` call and stamps the elapsed milliseconds on the attempt (`Attempt.LatencyMs`, `latency_ms` on the API object). One wrap point covers every destination type, and it is the call `DeliveryTimeout` already bounds, so a timed-out attempt records roughly the timeout rather than nothing.

- Success and failed attempts both get a value (a `timeout` or `connection_refused` failure records the time until the error). `filters[status]=success` gives success-only latency, composing with the existing API instead of needing a separate measure.
- Attempts that never reach `Publish` (publisher resolution / validation failure) have `latency_ms: null`.
- Nacked attempts are not recorded today and that does not change.

**Persist** — one nullable integer column per SQL store, no index (it is only ever aggregated):

```sql
-- clickhouse/000004_attempt_latency
ALTER TABLE {deployment_prefix}attempts ADD COLUMN latency_ms Nullable(UInt32);
-- postgres/000011_attempt_latency
ALTER TABLE attempts ADD COLUMN latency_ms integer;
```

Nullable so that attempts written before the migration are skipped by the aggregates instead of dragging averages toward zero, and so the API can return `null` for them instead of a fake `0`. Carried through insert / list / retrieve in `chlogstore`, `pglogstore` and `memlogstore`.

**Expose** — four new measures on `GET /metrics/attempts`, in milliseconds:

| Measure | ClickHouse | Postgres | Memory |
|---|---|---|---|
| `avg_latency` | `avg(latency_ms)` | `AVG(latency_ms)` | mean |
| `p50_latency` / `p95_latency` / `p99_latency` | one `quantilesTDigest(0.5, 0.95, 0.99)(latency_ms)` state per group | `percentile_cont(p) WITHIN GROUP (ORDER BY latency_ms)` | linear interpolation between closest ranks (same definition as `percentile_cont`) |

They inherit every dimension, filter and granularity. For a bucket in which no attempt recorded a latency (including zero-filled empty buckets) they are `null`, not `0` — a 0ms point on a response-time chart would be a lie. This is the one deliberate deviation from the zero-fill convention the other measures use, flagged in the discussion and agreed there.

**Spec / docs** — `docs/apis/openapi.yaml` (measures enum + descriptions + response example, `latency_ms` on `Attempt` and its examples) and `docs/content/features/metrics.mdoc`. `speakeasy lint openapi` is clean (0 errors; the one warning is pre-existing). No `sdks/*` edits.

**Bench** — `cmd/bench/metrics` seeds `latency_ms` and gains four latency cases. The seed scripts also needed a fix to run at all on `main`: they still inserted `events.destination_id`, which was replaced by `matched_destination_ids` in migration 000002 (CH) / 000007 (PG).

## ClickHouse performance

The concern raised in the discussion. `quantilesTDigest` is single-pass and bounded-memory (a t-digest of a few KB per group), so a percentile query costs about the same as the existing `avg(attempt_number)`, and all three percentiles share one aggregate state regardless of how many are requested. `quantileExact` was avoided because it materializes every value per group.

Numbers from `cmd/bench/metrics` on the 10M-event / 12.6M-attempt seed, single laptop (Apple M-series, Docker), `-benchtime=3x`:

ClickHouse 24.10, 12.6M attempts (`p50` of the seed ≈ 273ms, `p95` ≈ 501ms, `p99` = 5000ms):

| Case | Query | ns/op |
|---|---|---|
| CountAll (existing) | `count` over the month | 1,345,853,667 |
| FailedRate (existing) | `failed_rate` over the month | 681,751,181 |
| **AvgLatencyAll** | `avg_latency` over the month | 317,576,819 |
| **LatencyPercentilesAll** | `p50/p95/p99_latency` over the month | 466,626,861 |
| **LatencyPercentilesByDestinationHourly** | `count, avg_latency, p95_latency` by `destination_id`, 1h buckets, 1 day | 75,478,417 |
| **LatencyPercentilesSingleDestinationDaily** | `count, error_rate, avg, p50, p95, p99` for one destination, 1d buckets, month | 544,611,820 |
| Hourly_1Day (existing) | `count`, 1h buckets, 1 day | 30,638,597 |
| Daily_1Month (existing) | `count`, 1d buckets, month | 1,327,278,278 |
| AllMeasures (existing, pre-latency set) | every existing measure over the month | 4,316,119,264 |

The latency aggregates are cheaper than the existing `uniqExact(attempt_id)`-based counts on the same range, so adding them to a dashboard query does not move its cost class.

Postgres 16, same 12.6M attempts, for reference (exact `percentile_cont`):

| Case | ns/op |
|---|---|
| CountAll (existing) | 5,753,929,430 |
| FailedRate (existing) | 2,960,711,139 |
| **AvgLatencyAll** | 3,105,145,236 |
| **LatencyPercentilesAll** | 3,850,318,972 |
| **LatencyPercentilesByDestinationHourly** | 458,647,722 |
| **LatencyPercentilesSingleDestinationDaily** | 33,966,514 |
| Hourly_1Day (existing) | 77,596,583 |
| Daily_1Month (existing) | 3,145,644,764 |
| AllMeasures (existing, pre-latency set) | 3,422,379,903 |

Duplicates from unmerged `ReplacingMergeTree` parts carry identical values and only negligibly affect avg/percentiles, the same reasoning as the existing comment on `avg(attempt_number)`.

## Testing

- `drivertest` conformance, run against ClickHouse 24.10, Postgres 16 and the in-memory driver: the metrics dataset now assigns `100 + (i%10)*50` ms to non-manual attempts and no latency to the 30 manual ones, so the expected avg (300), p50 (300), p95/p99 (500) are only right if NULLs are skipped rather than counted as 0. Also asserts `null` for an all-NULL group, for a tenant with no recorded latency, and for empty buckets, and that latency round-trips through insert / list / retrieve with nil staying nil. Percentiles are pinned exactly on PG/mem and within 25ms on ClickHouse (values are 50ms apart, so that still resolves to one expected value).
- `destregistry`: latency is set on success, set on a failed attempt that carries a delivery, and unset when the publisher could not be resolved.
- `apirouter`: the four measures on `/metrics/attempts` (values, `null` when nothing recorded, `null` in empty buckets, 400 on unknown measure unchanged), and `latency_ms` on the attempt object including the `null` case.
- `go test ./...` green; the logstore, apirouter and destregistry suites green with `TESTINFRA=1`.
- End to end on a locally built `outpost-server` (ClickHouse log store, RabbitMQ): a destination pointed at a mock that sleeps 250–550ms and returns 503 for one event. Attempts came back with `latency_ms` of 259–562 and the 503 with its own latency; `/metrics/attempts` returned `avg_latency: 406.75, p50: 356, p95: 562, p99: 562, error_rate: 0.125` for the destination, `filters[status]=success` excluded the failure, and 30-minute buckets with no data returned `null`.

## Notes for review

- Measure names follow the existing `avg_attempt_number` prefix style. Happy to rename if you prefer the `avg_delivery_latency` form mentioned in the thread.
- Sub-millisecond replies (a localhost mock) round down to `0`; that is inherent to millisecond storage and not worth a wider column.
- `internal/logstore/chlogstore/README.md` still documents the pre-migration single `event_log` table and wasn't touched; it is stale independently of this change.

Closes the proposal in #1074.


internal/logstore/chlogstore/metrics.go
```diff
@@ -379,6 +383,20 @@ func (s *logStoreImpl) QueryAttemptMetrics(ctx context.Context, req driver.Metri
...
+			selectExprs = append(selectExprs, "avg(latency_ms)")
+			order = append(order, sfAvgLatency)
+		case "p50_latency", "p95_latency", "p99_latency":
+			if wantLatencyQuantiles {
+				continue
+			}
+			wantLatencyQuantiles = true
+			// One t-digest state per group serves all three percentiles;
+			// quantileExact would materialize every value per group.
+			selectExprs = append(selectExprs, "quantilesTDigest(0.5, 0.95, 0.99)(latency_ms)")
```
alexluong: nit: have you looked at `quantilesTiming`? It's built for millisecond latencies, near-exact under 30s, and deterministic. t-digest results can shift slightly between runs of the same query, which shows up as jitter on a chart. If it benchmarks about the same, I'd lean toward it.

internal/models/entities.go
```diff
@@ -138,6 +138,9 @@ type Attempt struct {
 	Time            time.Time              `json:"time"`
 	Code            string                 `json:"code"`
 	ResponseData    map[string]interface{} `json:"response_data"`
+	// Provider response time in ms, measured around Publish. Nil when the
+	// attempt never reached the provider or predates latency tracking.
+	LatencyMs *int64 `json:"latency_ms"`
```
alexluong, line 143: Operator event payloads (`attempt.success`, `attempt.failed`, `exhausted_retries`, etc.) include `*models.Attempt`, so `latency_ms` now appears there too. Fine by me, but could you add it to the example in `docs/content/features/operator-events.mdoc`?

internal/destregistry/registry_test.go
```diff
@@ -700,6 +704,80 @@ func TestPublishEventTimeout(t *testing.T) {
...
+		require.NoError(t, err)
+		provider.publishDelay = delay
+		require.NoError(t, registry.RegisterProvider("test", provider))
+
+		attempt, err := registry.PublishEvent(context.Background(), &models.Destination{Type: "test"}, &models.Event{})
+		require.NoError(t, err)
+		require.NotNil(t, attempt)
+		require.NotNil(t, attempt.LatencyMs, "latency must be recorded on success")
+		assert.GreaterOrEqual(t, *attempt.LatencyMs, delay.Milliseconds())
+		assert.Less(t, *attempt.LatencyMs, (10 * delay).Milliseconds(), "latency should measure the publish call, not the whole test")
```
alexluong: nit: this upper bound (500ms) could fail on a busy CI runner, since the test runs in parallel. I'd drop it or raise it a lot. The lower bound already shows the publish call is timed.

alexluong (commented):
Thanks, this is clean and follows the thread closely. Also appreciate the seed script fix.

One ask on ClickHouse perf before approving. The benchmarks compare latency queries to `count`, which is slow mainly because of `uniqExact(attempt_id)`. What I'd like to see is the added cost of latency on a query a dashboard already runs. For example, `count, error_rate` by `destination_id` with 1h buckets, then the same query plus `avg_latency, p95_latency`. Could you add that pair to `cmd/bench/metrics`, ideally over both a day and a month?

kunwar-axilio:
Thanks for the review. All four addressed, rebased on `main`:

**Marginal cost of latency on a dashboard query** 
Added `Dashboard_*` / `DashboardWithLatency_*` pairs to `cmd/bench/metrics` (`count, error_rate` by `destination_id`, with vs without `avg_latency, p95_latency`), hourly over a day and a month plus daily over a month. Numbers below are three alternating runs of the exact SQL the driver builds, on the 12.6M-attempt seed.

| Shape | without latency (s) | with latency (s) |
|---|---|---|
| 1h buckets, 1 day | 0.055 / 0.047 / 0.047 | 0.056 / 0.052 / 0.050 |
| 1h buckets, 1 month | 1.474 / 1.456 / 1.400 | 1.478 / 1.484 / 1.720 |
| 1d buckets, 1 month | 1.365 / 1.389 / 1.395 | 1.419 / 1.478 / 1.424 |

So ~+3ms on the day shape, +2–5% on the month shapes. 

**quantilesTiming** 
Switched that as well. On the same grouped month query it's ~15% faster than t-digest (1.51s vs 1.80s, three runs each), the `drivertest` percentile pins now come out exact on ClickHouse, and it's deterministic. The values are bounded by `DeliveryTimeout`, so the 30s ceiling should never come up.

**Operator events** 
`latency_ms` added to the `attempt.success` / `attempt.failed` example in `operator-events.mdoc`.

**Upper bound in the registry test**
Dropped; the lower bound is the assertion that matters.

PR body updated with the new numbers.

alexluong (approved):
Thanks for the prompt resolution! LGTM.

